PR #7, opened 2025-07-17
Fix isInsideModule check

robertjpayne:
* Ensure src and dist folders are checked from local development structure


robertjpayne:
I'm not sure this one is 100% ready to go and honestly the two development checks look a bit suss if they are fit for purpose. If they are truly only meant to be there for development perhaps they should calculate the dirname of the babel plugin script and formulate an check path based on that directly?

marklawlor:
This has been reworked on main

PR #126, opened 2025-08-27
feat: react-native-svg support

marklawlor:


danstepanov:

Since this is Mark's own PR, these are just notes for whenever it gets picked back up. SVG support is in the 5.0 scope, but this branch has drifted. It conflicts with main, wraps only Circle, and the single test is a skip stub. The compiler half already landed on main with passing vendor tests for fill, stroke and stroke width, so what remains is the component wrapper set plus a metro resolver entry for react-native-svg, following the safe area context native and web split, with react-native-svg as an optional peer. It would also be the natural place to fix the styled() stack overflow in nativewind#1820.

PR #221, opened 2025-10-22
fix: add ref forwarding support to styled()

tyrauber:
- fix ref forwarding issues preventing SVG and native components from receiving props
- maintain full backward compatibility with existing styled() configurations
- add test coverage for ref forwarding

Fixes issue where styled() didn't work with react-native-svg and expo-image components due to missing ref forwarding.

Users can now manually configure nativeStyleMapping for native components and it will work correctly.

### DIFF
--- a/src/native/api.tsx
+++ b/src/native/api.tsx
@@ -1,11 +1,5 @@
 /* eslint-disable  */
-import {
-  createElement,
-  forwardRef,
-  useContext,
-  useState,
-  type ComponentType,
-} from "react";
+import { forwardRef, useContext, useState, type ComponentType } from "react";
 import { Appearance } from "react-native";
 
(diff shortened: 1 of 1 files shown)


tyrauber:
Actually, this might not resolve the issue. 

It seems like only the global assignment fails for native component props, direct assignment works.

```
const StyledImage = styled(Image);

export default () => {
  return (
    <View className="flex-1 justify-center items-center">
      <StyledImage
        className="w-[200px] h-[200px]"
        source={require('@/assets/images/whatever.png')}
        contentFit="cover"
        transition={1000}
      />
    </View>
  );
};
```

The above works with react-native-css 3.0.0.  It is only the global assignment as detailed in the nativewind  v5 documentation that fails for native prop components.

Marking as draft, until I can debug further.  The above is a work around for expo-image and svgs.

PR #227, opened 2025-11-09
fix: correct parameter order in function call

sascha-kaleidoscode:
Fix: Correct parameter order in useCssElement call (web)

This fixes a bug in the styled() helper where the arguments to useCssElement were passed in the wrong order. The function expects (component, props, mapping), but we were passing (component, mapping, props) by mistake.

This meant the style mapping was being treated as props, which caused styles to be applied incorrectly. The parameter order is now fixed so everything behaves as intended.


danstepanov:
Thanks for catching this, just waiting for CI to pass and will merge

Issue #232, opened 2025-11-24
[V5] - TextInput crashes when using "text-center" or "text-right" or "text-left" in NativeWind v5 (react-native-css: path.split is undefined)


MohammedAbidNafi:
**Describe the bug**
When adding the text-center, text-left, text-right class to a React Native TextInput, the app crashes with Render Error the following is the error:

`path.split is not a function (it is undefined)`

<img width="400" height="870" alt="Image" src="https://github.com/user-attachments/assets/4e1d00a4-f7ec-4851-8301-3ab94aa8ab7d" />

The error originates in react-native-css at src/native/styles/index.ts in line 250

It appears path is undefined (or non-string) during runtime parsing of the class path. This crash happens when text-center, text-left, text-right  is applied to a TextInput. Other components appear unaffected

**Reproduction**
I reproduced using the recommended template and then upgrading to NativeWind v5:

`npx rn-new@latest --nativewind --expo-router`

Migrated to NativeWind v5

And added this minimal component:

`
import { TextInput } from 'react-native';

export default function Home() {
  return (
        <"TextInput"
          placeholder="Type here..."
          className="mb-4 rounded-md border border-gray-300 p-2 text-center"
        />
  );
}
`

**Expected behavior**
Applying text-center or text-left or text-right to TextInput should center the input text without throwing an exception.

danstepanov:
I've created a minimal repro here and moved this issue to the react-native-css repo https://github.com/danstepanov/text-alignment-bug

PR #251, opened 2025-12-21
Fixing Issue with text classes from TW 5

mo-fouad:
When using Variants with text-centre, text-left, or text-right, we have a Bug as it shows [here](https://github.com/nativewind/react-native-css/issues/232)

### DIFF
--- a/src/components/TextInput.tsx
+++ b/src/components/TextInput.tsx
@@ -12,7 +12,7 @@ const mapping: StyledConfiguration<typeof RNTextInput> = {
   className: {
     target: "style",
     nativeStyleMapping: {
-      textAlign: true,
+      textAlign: "textAlign",
     },
   },
 };


danstepanov:
Thanks for opening this, I'll take a look this week

danstepanov (changes requested):
Hey @mo-fouad, thanks for the contribution. This one-liner would fix this specific case but I actually think this is a symptom of a bigger issue with `nativeStyleMapping()`. There are similar situations in which a similar error would occur. Rather than merge this right now, I'd like to implement a fix for [nativeStyleMapping()](https://github.com/nativewind/react-native-css/blob/008d479253365745acedef5a9602a7382f991597/src/native/styles/index.ts#L420).

The more general purpose fix would use the key name as the path in [this for loop](https://github.com/nativewind/react-native-css/blob/008d479253365745acedef5a9602a7382f991597/src/native/styles/index.ts#L453).

I was thinking something like `const path = pathOrTrue === true ? key : pathOrTrue;`

danstepanov:

The crash was real when this was opened, but the root cause was fixed on main by #323, which normalizes the boolean shorthand at config creation, and #232 is closed. Against current main this change is a behavioral no op, so it can close as superseded.

PR #277, opened 2026-01-11
feat: add inset shadow parsing support

YevheniiKotyrlo:
## Problem

Tailwind CSS v4 shadow and ring utilities don't work on React Native. Two root causes:

1. **Inset shadows** — The box-shadow shorthand handler doesn't recognize the `inset` keyword, so shadows arriving through CSS variables at runtime fail to parse. **(Fixed here)**
2. **Missing `@property` defaults** — Tailwind v4's `@property` declarations aren't parsed, so variables like `--tw-ring-shadow` have no initial values. **(Fixed in #284, stacked on this PR)**

This is **PR 1/2**. PR #284 builds directly on this branch.

### `src/native/styles/shorthands/box-shadow.ts`
- Implemented `inset` field descriptor with literal string matching
- Added 4 inset shorthand patterns: with/without spread, with/without color, color-first
- Added `normalizeInsetValue()` to convert string `"inset"` to boolean `true` for React Native's `boxShadow` prop

### `src/__tests__/native/box-shadow.test.tsx`
- 4 compile-time tests: basic inset, color-first, inherited color, mixed inset+regular
- 2 runtime tests: inset via CSS variable, inset via CSS variable without explicit spread
- 1 CSS variable resolution test with strict assertions

## Review notes

- Per review feedback, all Tailwind-specific root variable defaults were removed from this PR — those are handled generically via `@property` support in #284
- All tests use `toStrictEqual` with exact expected values
- Two files changed, +192/−3 lines

### DIFF
--- a/src/__tests__/native/box-shadow.test.tsx
+++ b/src/__tests__/native/box-shadow.test.tsx
@@ -93,3 +93,169 @@ test("shadow values - multiple nested variables", () => {
     ],
   });
 });
+
+test("inset shadow - basic", () => {
+  registerCSS(`
+    .test { box-shadow: inset 0 2px 4px 0 #000; }
+  `);
+
+  render(<View testID={testID} className="test" />);
+  const component = screen.getByTestId(testID);
+
+  expect(component.props.style).toStrictEqual({
+    boxShadow: [
+      {
+        inset: true,
+        offsetX: 0,
+        offsetY: 2,
+        blurRadius: 4,
+        spreadDistance: 0,
+        color: "#000",
+      },
+    ],
+  });
+});
+
+test("inset shadow - with color first", () => {
+  registerCSS(`
+    .test { box-shadow: inset #fb2c36 0 0 24px 0; }
+  `);
+
+  render(<View testID={testID} className="test" />);
+  const component = screen.getByTestId(testID);
+
+  expect(component.props.style).toStrictEqual({
+    boxShadow: [
+      {
+        inset: true,
+        color: "#fb2c36",
+        offsetX: 0,
+        offsetY: 0,
+        blurRadius: 24,
+        spreadDistance: 0,
+      },
+    ],
+  });
+});
+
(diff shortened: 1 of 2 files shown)


danstepanov (changes requested):
Thanks for the contribution. The inset shadow parsing is solid. I have concerns about one part though.

1. Inset shadow parsing (`box-shadow.ts`) looks good. The new shorthand patterns and `normalizeInsetValue` correctly handle inset shadows coming through CSS variables at runtime. Statically-written inset shadows are already handled at compile time by lightningcss and the compiler, so this specifically covers the runtime/variable path, which is what matters for Tailwind's generated output. Minor gap: `[inset, offsetX, offsetY, blurRadius, color]` isn't covered, though Tailwind doesn't generate that pattern so not a blocker.

2. Root variable defaults (`root.ts`) needs to move. Hardcoding `--tw-shadow`, `--tw-inset-shadow`, `--tw-ring-shadow`, etc. in react-native-css couples the generic CSS engine to Tailwind's internal variable names. react-native-css is framework-agnostic and shouldn't have knowledge of Tailwind-specific variables. These defaults belong in [nativewind](github.com/nativewind/nativewind) (e.g., in `nativewind/theme.css` as `:root` variable declarations). That keeps the separation clean. The architectural intention of v5 is that Nativewind adapts Tailwind for RN while react-native-css handles generic CSS. 

Also, `rootVariables("tw-shadow-color").set([["initial"]])` seems problematic. What does `"initial"` resolve to at runtime? It's not a value react-native-css knows how to interpret.

3. The inset shadow tests are good. The "Tailwind v4 shadow variables" test should assert more than `.toBeDefined()`. Since `#0000` shadows get filtered by `omitTransparentShadows`, the test should verify what the resulting array actually contains.

TLDR: Could you split this into two PRs? The inset shadow changes to `box-shadow.ts` and tests are ready to merge. The root variable defaults need to move to nativewind.

PR #284, opened 2026-02-24
feat: add CSS @property rule support

YevheniiKotyrlo:
## Problem

Tailwind CSS v4 uses ~53 `@property` declarations to register custom property initial values (e.g., `--tw-shadow`, `--tw-ring-shadow`). Previously, react-native-css ignored `@property` rules entirely and faked `-moz-orient` support to trigger Tailwind's `@supports` fallback path. This fallback injected universal variable defaults which interfered with class-level overrides, causing utilities like `ring-2` to produce `boxShadow: []`.

The `@property` approach is generic and framework-agnostic — this is the direction suggested during the #277 review.

### `src/compiler/compiler.ts`
Parse `@property` rules, extract `initial-value` into root variables via `extractPropertyRule` / `parsePropertyInitialValue`. Handles length, number, percentage, color, angle, token-list, custom-ident, and repeated component types.

### `src/compiler/stylesheet.ts`
Add `addRootVariable()` public method following the existing variable storage pattern.

### `src/compiler/supports.ts`
Remove the `-moz-orient` hack (no longer needed).

### `src/native/styles/shorthands/_handler.ts`
Allow `StyleFunction` values (calc, var) to pass type checks in the shorthand handler. These are unresolved at pattern-match time but resolve to the correct type at runtime — rejecting them would break variable-based shadows entirely.

### `src/native/styles/shorthands/box-shadow.ts`
Fix color type from `"string"` to `"color"` to accept platform color objects (e.g., `PlatformColor` from `currentcolor` resolution).

### Tests
- 12 compiler tests for `@property` parsing (including single-child and multi-child `repeated` types)
- 2 runtime tests for box-shadow @property defaults and currentcolor matching

## Verification

- `yarn typecheck` — pass
- `yarn lint` — pass
- `yarn build` — pass (no unstaged files)
- `yarn test` — 990 passed, 3 failed (pre-existing babel plugin failures, same on `main`)

## Related

- Companion to #277 (now merged) — implements the `@property` approach suggested during that review
- #290 (merged) — `@types/debug` dependency fix in the same repo

### DIFF
--- a/src/__tests__/compiler/property.test.ts
+++ b/src/__tests__/compiler/property.test.ts
@@ -1,0 +1,226 @@
+import { compile } from "react-native-css/compiler";
+
+test("@property with length initial value", () => {
+  const compiled = compile(`
+@property --tw-translate-x {
+  syntax: "<length-percentage>";
+  inherits: false;
+  initial-value: 0px;
+}
+`);
+
+  const result = compiled.stylesheet();
+  expect(result.vr).toBeDefined();
+
+  const vrMap = new Map(result.vr);
+  expect(vrMap.has("tw-translate-x")).toBe(true);
+  expect(vrMap.get("tw-translate-x")).toStrictEqual([[0]]);
+});
+
+test("@property without initial value is skipped", () => {
+  const compiled = compile(`
+@property --tw-ring-color {
+  syntax: "*";
+  inherits: false;
+}
+`);
+
+  const result = compiled.stylesheet();
+  expect(result.vr).toBeUndefined();
+});
+
+test("@property with number initial value", () => {
+  const compiled = compile(`
+@property --tw-backdrop-opacity {
+  syntax: "<number>";
+  inherits: false;
+  initial-value: 1;
+}
+`);
+
+  const result = compiled.stylesheet();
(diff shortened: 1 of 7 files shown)


YevheniiKotyrlo:
Rebased on latest `main` (`b0b35b1`, includes merged #277). Conflicts resolved — single clean commit.

Addressed both review comments:

1. **`isStyleFunction` comment expanded** — now explains the tradeoff: accepting unresolved functions makes pattern matching less strict when variables are involved, but rejecting them would break variable-based shadows entirely (e.g. `box-shadow: var(--shadow)` where `--shadow` resolves to `0 4px 6px -1px #000`).

2. **`repeated` single-child unwrap** — added two tests proving the behavior:
   - `@property with repeated single-child unwraps to scalar` — `<length>+` with `10px` produces `[[10]]` (same shape as direct `<length>`)
   - `@property with repeated multi-child preserves array` — `<length>+` with `10px 20px` produces `[[[10, 20]]]`

   Also expanded the inline comment to clarify why the unwrap exists.

All quality gates pass (typecheck, lint, build with no unstaged files, test — 990 passed, 3 pre-existing babel failures same on `main`).

Issue #233, opened 2025-10-15
Bug: Cannot set className and style at the same time in v5


beebls:
**Describe the bug**
In web versions of tailwind, and Nativewind v4, setting both className and style on an element would combine the two styles, with duplicate definitions having the style tag's value overwrite the class.

In Nativewind v5, this no longer works. Setting only className or only style works, but when you set both, the className property is completely ignored, with only the values from style applying.

**Reproduction**
Please see our contribution guide: https://github.com/nativewind/nativewind/blob/main/contributing.md#opening-an-issue

Below is a github repo that contains two rn-new boilerplate projects, one using nativewind 4 and one nativewind 5. A demo code snippet in App.tsx demonstrates styling with only className, styling with only the style prop, and a combination of the two.
https://github.com/beebls/nativewind-issue-demo

**Expected behavior**
Using the below code snippet on nativewind 4 and 5 yields different results.
```ts
import { View } from 'react-native';
import './global.css';

export default function App() {
  return (
    <View className="h-full w-full items-center justify-center bg-red-800">
      <View className="bg-blue-400 p-4" style={{ borderWidth: 10 }}>
        <View style={{ backgroundColor: 'white', width: 100, height: 100 }}></View>
      </View>
    </View>
  );
}
```
<img width="913" height="1018" alt="Image" src="https://github.com/user-attachments/assets/07da29b2-f1f9-4773-a2f2-0a7a38a8681f" />
Notice how the border from the style property is applied on both, but the background color and padding are only working on nativewind 4.

**Additional context**
I've observed this both in expo go and in dev client apps.

Additionally, this happens even if you set `style={{}}` or `style={undefined}`

xenonwellz:
We all need v3.0.2

WorkAttackDev:
Any feedback! it's taking too long!

monyasau:
Nativewind has been just unprediatable lately, first i have to wrap some components with styled before it works, then you cant combine styles anymore, i think there should be a way or some config for choosing to subscribe to depreciated methods/systems since it's literally just styling

18jad:
No way this is not fixed yet. How such a big package ignored a whole community issue and kept shipping new features without fixing an actually problem that might make their users migrate to a new package because of it.

PR #291, opened 2026-03-04
fix: prevent style={undefined} from destroying computed styles in non-[style] targets

YevheniiKotyrlo:
## Summary

Fixes remaining `Object.assign` calls in `deepMergeConfig` that allow `style={undefined}` to destroy computed NativeWind styles on non-`["style"]` targets.

## Problem

PR #224 fixed the most common code path (`config.target = ["style"]`), but two `Object.assign({}, left, right)` calls remain unfixed at [lines 362](https://github.com/nativewind/react-native-css/blob/b0b35b1/src/native/styles/index.ts#L362) and [365](https://github.com/nativewind/react-native-css/blob/b0b35b1/src/native/styles/index.ts#L365).

`Object.assign` copies all enumerable own properties from `right`, including those with value `undefined`. When a component passes `contentContainerStyle={undefined}` alongside `contentContainerClassName`, the computed styles from `left` are overwritten with `undefined`.

**Affected components:**
- `ScrollView` — `contentContainerClassName` + `contentContainerStyle={undefined}`
- `FlatList` — `contentContainerClassName` / `columnWrapperClassName` + `*Style={undefined}`
- `VirtualizedList` — `contentContainerClassName` + `contentContainerStyle={undefined}`
- Any custom `styled()` with non-default target mappings (e.g. `{ className: { target: "style" } }`)

This is a common pattern in reusable components where style props are optional:

```tsx
function MyScrollView({ contentContainerStyle, ...props }: ScrollViewProps) {
  return (
    <ScrollView
      contentContainerClassName="p-4"
      contentContainerStyle={contentContainerStyle}
      {...props}
    />
  );
}
// <MyScrollView /> — contentContainerStyle is implicitly undefined, className styles lost
```

Native-only — web uses array concatenation (no `Object.assign`).

## Solution

Added a `mergeDefinedProps` helper that replaces both `Object.assign({}, left, right)` calls. It spreads `left` into a new object and iterates `right` via `for-in`, skipping keys with `undefined` values.

```typescript
function mergeDefinedProps(
  left: Record<string, any> | undefined,
  right: Record<string, any>,
) {
  const result = left ? { ...left } : {};
  for (const key in right) {
    if (right[key] !== undefined) {
      result[key] = right[key];
    }
  }
  return result;
}
```

This covers all `config.target` types (array, string, false) at all recursion depths, and still allows explicit `null` values to override (only `undefined` is skipped). Uses `for-in` instead of `Object.keys()` to avoid intermediate array allocation. Both `left` and `right` are always plain object literals (JSX props or `calculateProps` output) — no prototype chain concerns.

### Performance

Benchmarked 15 implementation variants of `mergeDefinedProps` in isolation (2M iterations, best of 5 runs) across 6 scenarios matching real component renders, then validated the winner in the full `deepMergeConfig` context (1M iterations each):

| Scenario | Original | Fixed | Ratio |
|---|---|---|---|
| ScrollView + `style={undefined}` (the bug) | 129ms | 113ms | **0.87x** (13% faster) |
| ScrollView + valid style | 125ms | 122ms | **0.97x** (same) |
| FlatList + many props + undefined styles | 320ms | 302ms | **0.94x** (6% faster) |
| Custom `styled()` + `style={undefined}` | 172ms | 161ms | **0.94x** (6% faster) |

The fix is **faster than the original** `Object.assign` across all real-world scenarios.

## Verification

- `yarn typecheck` — pass
- `yarn lint` — pass
- `yarn build` — pass (no unstaged files)
- `yarn test` — 984 passed, 3 failed (pre-existing babel plugin failures, same on `main`)

Added 8 test cases to `className-with-style.test.tsx` covering all `config.target` types:
- **Path A** (`["style"]`): `style={undefined}` and `style={null}` preserve computed styles
- **Path B** (`["contentContainerStyle"]`): `ScrollView` and `FlatList` with `contentContainerStyle={undefined}` preserve computed styles; valid `contentContainerStyle` merges correctly; omitted prop preserves styles
- **Path B** (string target): custom `styled()` with `{ target: "style" }` preserves styles with `style={undefined}`
- **Real-world**: optional `contentContainerStyle` prop forwarding (implicitly `undefined`) preserves computed styles

## Reproduction

Visual reproduction repo with before/after screenshots: [nativewind-style-undefined-repro](https://github.com/YevheniiKotyrlo/nativewind-style-undefined-repro)

<table>
<tr><th>Before (3.0.4 unpatched)</th><th>After (patched)</th></tr>
<tr>
<td></td>
<td></td>
</tr>
</table>

## Related

- Completes the fix started in #224
- Related to #233 (last comment reports `style={undefined}` still breaking after #224)

### DIFF
--- a/src/__tests__/native/className-with-style.test.tsx
+++ b/src/__tests__/native/className-with-style.test.tsx
@@ -1,7 +1,17 @@
+import { View as RNView } from "react-native";
+
 import { render } from "@testing-library/react-native";
+import { copyComponentProperties } from "react-native-css/components/copyComponentProperties";
+import { FlatList } from "react-native-css/components/FlatList";
+import { ScrollView } from "react-native-css/components/ScrollView";
 import { Text } from "react-native-css/components/Text";
 import { View } from "react-native-css/components/View";
 import { registerCSS, testID } from "react-native-css/jest";
+import {
+  useCssElement,
+  type StyledConfiguration,
+  type StyledProps,
+} from "react-native-css/native";
 
 test("className with inline style props should coexist when different properties", () => {
   registerCSS(`.text-red { color: red; }`);
@@ -11,8 +21,8 @@ test("className with inline style props should coexist when different properties
   ).getByTestId(testID);
 
   // Both className and style props should be applied as array
-  expect(component.props.style).toEqual([
-    { color: "#f00" }, // Changed from "red" to "#f00"
+  expect(component.props.style).toStrictEqual([
+    { color: "#f00" },
     { fontSize: 16 },
   ]);
 });
@@ -24,8 +34,8 @@ test("className with inline style props should favor inline when same property",
     <Text testID={testID} className="text-red" style={{ color: "blue" }} />,
   ).getByTestId(testID);
 
-  // When same property exists, inline style should win (not array)
-  expect(component.props.style).toEqual({ color: "blue" });
+  // When same property exists, inline style should win
+  expect(component.props.style).toStrictEqual({ color: "blue" });
 });
 
 test("only className should not create array", () => {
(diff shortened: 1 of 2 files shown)


YevheniiKotyrlo:
Done — rebased on latest `main` (`251cdfc`). All quality gates pass (typecheck, lint, build, tests).

PR #344, opened 2026-05-24
fix: avoid walking ref props in styled mappings

### LINKED ISSUE #343: TypeScript hangs when styled() wraps components with complex ref props
### Before submitting a new issue

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a supported version of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

`styled()` can cause TypeScript to hang when wrapping components whose props include complex `ref` types, such as `react-strict-dom` elements.

Minimal example:

```tsx
import { html } from "react-strict-dom";
import { styled } from "react-native-css";

export const StyledDiv = styled(html.div, { className: "style" });

export const element = <StyledDiv className="p-4" />;
```

### What I found

The issue appears to come from `StyledConfiguration` computing dot-notation style targets from:

```ts
DotNotation<ComponentProps<C>>
```

For `html.div`, `ComponentProps<C>` includes:

```ts
ref?: React.Ref<HTMLDivElement>
```

`DotNotation` then recursively walks into the `ref` type, then into `HTMLDivElement`, which expands into a large recursive DOM type graph.

I verified locally that:

- `ComponentProps<typeof html.div>` typechecks quickly.
- `DotNotation<Pick<Props, "style">>` typechecks quickly.
- `DotNotation<Pick<Props, "ref">>` hangs.
- `styled(html.div, { className: "style" })` hangs before the fix.

### Library version

3.0.7

### Environment info

```shell
Node: 24.13.0
TypeScript: 5.9.2
react-native-css: 3.0.7
react-strict-dom: 0.0.55
react: 19.1.0
react-native: 0.81.4
```

### Steps to reproduce

1. Clone the repro repository.
2. Install dependencies.
3. Run `npm run typecheck`.

### Reproducible example repository

https://github.com/emmanuelchucks/react-native-css-styled-rsd-ref-repro

### Possible fix

Using `PropsWithoutRef<ComponentProps<C>>` for style-mapping target inference avoids walking `ref`, while still preserving `ref` on the returned styled component props.

Conceptually:

```ts
DotNotation<PropsWithoutRef<ComponentProps<C>>>
```

instead of:

```ts
DotNotation<ComponentProps<C>>
```

emmanuelchucks:
## Summary

Fixes a TypeScript performance issue where `styled()` can hang when wrapping components whose props include complex `ref` types, such as `react-strict-dom` elements.

`StyledConfiguration` currently computes dot-notation style targets from `ComponentProps<C>`. For components like `html.div`, that includes `ref?: React.Ref<HTMLDivElement>`. `DotNotation` then recursively walks into DOM element types through `ref`, which can make TypeScript hang.

This changes style-mapping target inference to use `PropsWithoutRef<ComponentProps<C>>`, while preserving `ref` on the returned styled component props.

Closes #343.

## Reproduction

Minimal repro: https://github.com/emmanuelchucks/react-native-css-styled-rsd-ref-repro

```tsx
import { html } from "react-strict-dom";
import { styled } from "react-native-css";

export const StyledDiv = styled(html.div, { className: "style" });
export const element = <StyledDiv className="p-4" />;
```

## Test plan

Ran locally:

```bash
yarn typecheck
yarn lint
yarn test --maxWorkers=2
yarn build
yarn example expo export --platform web
```

Also verified the `react-strict-dom` repro typechecks after the change.

### DIFF
--- a/src/runtime.types.ts
+++ b/src/runtime.types.ts
@@ -6,6 +6,7 @@ import type {
   ComponentType,
   ForwardRefExoticComponent,
   FunctionComponent,
+  PropsWithoutRef,
   ReactElement,
 } from "react";
 import type {
@@ -84,15 +85,15 @@ interface StyledConfigurationObject<
 > {
   target: T;
   nativeStyleMapping?: T extends false
-    ? NativeStyleMapping<string, ComponentProps<C>>
+    ? NativeStyleMapping<string, PropsWithoutRef<ComponentProps<C>>>
     : NativeStyleMapping<
         ResolveDotPath<T, ComponentProps<C>>,
-        ComponentProps<C>
+        PropsWithoutRef<ComponentProps<C>>
       >;
   /** @deprecated Please use nativeStyleMapping */
   nativeStyleToProp?: NativeStyleMapping<
     ResolveDotPath<T, ComponentProps<C>>,
-    ComponentProps<C>
+    PropsWithoutRef<ComponentProps<C>>
   >;
 }
 
@@ -119,7 +120,7 @@ export type ReactComponent<P = any> =
   | ForwardRefExoticComponent<P>;
 
 export type ComponentPropsDotNotation<C extends ReactComponent> = DotNotation<
-  ComponentProps<C>
+  PropsWithoutRef<ComponentProps<C>>
 >;
 
 /********************************    Styles    ********************************/


danstepanov:

The diagnosis in #343 is right. DotNotation enumerates ComponentProps including ref, and a DOM typed ref drags the whole DOM type graph into the walk. PropsWithoutRef stops it at the source, ref still works on the returned component, and it's type level only with zero runtime change. Applies cleanly to main. Not required to ship latest but a safe one file fix for a confirmed bug, easy merge now or in the first 5.x patch. Unrelated note while I was in here, StyledConfigurationObject passes ResolveDotPath's arguments in swapped order and only typechecks because the generic collapses to any, worth a separate look.

PR #345, opened 2026-05-26
fix: apply effective rem to calc function using variables

Jin-Byun:
### Summary
Fix for a bug: Custom `effectiveRem` value not being applied in Tailwind V4.

- Tailwind V4 uses `calc` function with rem-based variables (e.g `--spacing: 0.25rem`) for dynamic values (e.g. `w-1 px-17 h-1.5`) and more. (e.g. `.rounded-4xl { border-radius: calc(var(--radius-xl) * 4}`)
- `Length` visitor in the compiler cannot update `calc` function with rem-based variable directly into `px` value.
- Update in the `Declaration` visitor using remToPx transformation similar to `Length` visitor catches and convert rem-based variables to px-based variables.

### Test Plan
- [x] Added test to check whether the `effectiveRem` set by `:root { font-size: 16px; }` is applied to `border-width: calc(var(--spacing) * 4)`
- [x] Tests Passed, typechecked and linted.

### DIFF
--- a/src/__tests__/compiler/compiler.test.tsx
+++ b/src/__tests__/compiler/compiler.test.tsx
@@ -79,7 +79,7 @@ test(":root CSS variables with media queries", () => {
 
 test.skip("removes unused CSS variables", () => {
   const compiled = compile(`
-    .test { 
+    .test {
       --blue: blue;
       --green: green;
       --red: red;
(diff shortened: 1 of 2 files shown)


danstepanov:

The bug is real and blocker level, rem inside variable values bypasses the first pass conversion so calc(var(--spacing) * 4) resolves against the default 14. But #336 fixes the same symptom more completely with the inlineRem write back plus the Metro options propagation fix, while this patch misses rem nested inside function tokens. I'd land #336 and close this one, ideally carrying its compiler test over.

PR #346, opened 2026-05-26
feat: add cursor CSS property support

adrianorocha-dev:
## Summary

Add compiler support for the CSS `cursor` property, enabling Tailwind classes like `cursor-pointer`, `cursor-auto`, and `cursor-default` to compile to the native `cursor` style prop.

This is targeted at **react-native-macos** and **react-native-windows**, which both support `cursor` as a native style property (via `StyleSheet.create` or inline styles). iOS and Android silently ignore the prop, so this change is safe cross-platform.

Currently, lightningcss parses `cursor` declarations correctly, but the compiler has no parser registered for it — the property is silently dropped with a warning. This PR fixes that.

Also fixes a duplicate `cursor-default` test in the Tailwind interactivity tests (replaced the second one with `cursor-pointer`).

## Implementation

The parser extracts `value.keyword` from the lightningcss `Cursor` AST (`{ images: CursorImage[], keyword: CursorKeyword }`). Only the keyword is used — cursor images (e.g. `cursor: url(...), pointer`) are not handled since React Native does not support custom cursor images.

Similar to `parseUserSelect` and `parsePointerEvents`.

## Test plan

- [x] Added compiler unit tests for `cursor: pointer;` and `cursor: auto;`
- [x] Updated Tailwind interactivity tests from expecting warnings to expecting style output
- [x] Fixed duplicate `cursor-default` test → `cursor-pointer`
- [x] All 1042 tests pass (`yarn test`)
- [x] Lint and typecheck pass (lefthook pre-commit)

### DIFF
--- a/src/__tests__/compiler/declarations.test.tsx
+++ b/src/__tests__/compiler/declarations.test.tsx
@@ -12,6 +12,8 @@ const tests = [
   ["rotate: x 3deg;", [{ d: [[[{}, "rotateX", "3deg"], "rotateX"]], s: [1, 1] }]],
   ["stroke-width: 1px;", [{ d: [[1, ["strokeWidth"]]], s: [1, 1] }]],
   ["stroke: black;", [{ d: [["#000", ["stroke"]]], s: [1, 1] }]],
+  ["cursor: pointer;", [{ d: [{ cursor: "pointer" }], s: [1, 1] }]],
+  ["cursor: auto;", [{ d: [{ cursor: "auto" }], s: [1, 1] }]],
 ] as const;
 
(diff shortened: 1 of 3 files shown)


danstepanov:

Applies cleanly to current main and the implementation looks right, the keyword lands in the static declaration record which is exactly what the test asserts, and fixing the duplicate cursor default test name was a nice catch. v4 never supported cursor, so this is new functionality rather than an upgrade blocker. I'd slot it for the first 5.x minor after latest ships. Optional thought, core RN only accepts auto and pointer, so validating against an allowed list the way parseUserSelect does might be worth it, though unknown values are ignored natively so pass through is fine too.

Issue #288, opened 2026-02-26
nativeStyleMapping crashes when value is boolean (path.split is not a function)


clancywist:
## Bug

`nativeStyleMapping()` in `dist/module/native/styles/index.js` crashes when a `nativeStyleMapping` config entry has a boolean value (`true`) instead of a string path.

## Reproduction

Use `text-center` (or any class that produces `textAlign`) on a `<TextInput>`:

```tsx
<TextInput className="text-center text-2xl text-white" />
```

**Error:**
```
TypeError: path.split is not a function (it is undefined)
  at index.js:329
```

## Root cause

`TextInput.js` declares:
```js
const mapping = {
  className: {
    target: "style",
    nativeStyleMapping: {
      textAlign: true  // <-- boolean, not a string
    }
  }
};
```

At line 322-329 of `dist/module/native/styles/index.js`:
```js
for (const [key, path] of Object.entries(config.nativeStyleMapping)) {
  // ...
  const tokens = path.split(".");  // path is `true`, not a string → crash
}
```

`Object.entries({ textAlign: true })` yields `["textAlign", true]`. Then `true.split(".")` throws.

The same pattern exists in `ActivityIndicator` (`color: true`), `ImageBackground` (`backgroundColor: true`), and `Button` (`color: true`).

## Suggested fix

```js
const resolvedPath = typeof path === "string" ? path : key;
const tokens = resolvedPath.split(".");
```

When `path` is `true`, fall back to `key` as the prop name (same-name mapping).

## Environment

- `react-native-css`: 3.0.4
- `nativewind`: 5.x
- Expo SDK: 55
- Platform: iOS (Expo Go simulator)

RobertOstermann:
Related to #232 and #251

danstepanov:
Verified this is fixed in `3.1.0-rc.0`. Boolean `nativeStyleMapping` values now map the prop correctly without the `path.split` crash. The same test fails on 3.0.7.

PR #349, opened 2026-06-09
fix: handle boolean shorthand in nativeStyleMapping

### LINKED ISSUE #348: nativeStyleMapping crashes with "true.split is not a function" when mapping value is the boolean shorthand (e.g. TextInput textAlign)
## Summary

`nativeStyleMapping()` in `src/native/styles/index.ts` assumes every mapping value is a dot-path **string** and calls `path.split(".")`. However, the `NativeStyleMapping` type — and the built-in components shipped by this package — allow the value to be the boolean shorthand `true` (meaning "map onto a native prop with the same name"). When such a mapped style is actually present, `(true).split` throws `TypeError: undefined is not a function` during render.

Same root cause as #232 / #288 — filing with a precise root-cause trace and a one-line fix that I've verified locally via a patch.

## Affected built-in components

Components that ship a boolean-shorthand `nativeStyleMapping` crash as soon as the mapped style is applied:

- `components/TextInput.tsx` → `{ textAlign: true }` — triggered by `text-right` / `text-center` / `text-left`
- `components/ImageBackground.tsx` → `{ backgroundColor: true }` — triggered by any `bg-*`

(`ActivityIndicator` / `Button` use `{ color: "color" }`, a string path, so they don't hit this.)

## Reproduction

```tsx
import { TextInput } from "react-native-css/components"; // or via nativewind v5
<TextInput className="text-right" value="x" onChangeText={() => {}} />
```

Render → `TypeError: undefined is not a function`.

Note it only crashes when the mapped style is actually present: when `textAlign` is absent, `styleValue === undefined` short-circuits via `continue`, which is why `text-left`-less inputs render fine and the crash looks intermittent.

## Root cause

`src/native/styles/index.ts`:

```ts
for (const [key, path] of Object.entries(config.nativeStyleMapping)) {
  const styleValue = source[key];
  delete source[key];
  if (styleValue === undefined) {
    continue;
  }
  let target = props;
  const tokens = path.split(".");   // path === true  ->  true.split is not a function
  ...
}
```

`useNativeCss.ts` casts `value.nativeStyleMapping as Record<string, string>`, but the runtime value is `{ textAlign: true }`, so `path` is the boolean `true`.

## Suggested fix

Treat `true` as the same-name shorthand (target prop named `key`), matching the `true | DotNotation` type:

```ts
const tokens = (typeof path === "string" ? path : key).split(".");
```

## Environment

- react-native-css 3.0.7 (latest)
- nativewind 5.0.0-preview.3
- react-native 0.85.3 / React 19.2.3
- Hermes, iOS

Related: #232, #288, #251

telanflow:
## Summary

`nativeStyleMapping()` in `src/native/styles/index.ts` calls `path.split(".")` on every mapping value, but a value may be the boolean shorthand `true` — the same-name mapping permitted by the `NativeStyleMapping` type (`true | DotNotation`). When the mapped style is actually present, `(true).split` throws `TypeError: undefined is not a function`, crashing render.

Fixes #348 (same root cause as #232 / #288).

## Affected built-in components

Components that ship the boolean shorthand crash as soon as the mapped style is applied:

- `components/TextInput.tsx` → `{ textAlign: true }` — triggered by `text-right` / `text-center` / `text-left`
- `components/ImageBackground.tsx` → `{ backgroundColor: true }` — triggered by any `bg-*`

(`ActivityIndicator` / `Button` use `{ color: "color" }`, a string path, so they're unaffected.)

It only crashes when the mapped style is present: when absent, `styleValue === undefined` short-circuits via `continue`, which is why the crash looks intermittent.

## Fix

Treat `true` as the same-name shorthand (map onto a prop named `key`):

```ts
const tokens = (typeof path === "string" ? path : key).split(".");
```

## Test

Added a regression test in `src/__tests__/native/components.test.tsx` (using the `target: "style"` + `{ backgroundColor: true }` shorthand). It fails on `main` with `TypeError: path.split is not a function` and passes with this change.

`yarn test`, `yarn typecheck`, and `yarn lint` all pass.

### DIFF
--- a/src/__tests__/native/components.test.tsx
+++ b/src/__tests__/native/components.test.tsx
@@ -1,4 +1,9 @@
-import { Button as RNButton, type ButtonProps } from "react-native";
+import {
+  Button as RNButton,
+  View as RNView,
+  type ButtonProps,
+  type ViewProps,
+} from "react-native";
 
 import { render } from "@testing-library/react-native";
 import { copyComponentProperties } from "react-native-css/components/copyComponentProperties";
@@ -53,4 +58,31 @@ test("Component preserves props when mapping specifies 'target: false'", () => {
   expect(titleElement.props.style).toBeInstanceOf(Array);
   expect(titleElement.props.style).toHaveLength(2);
   expect(titleElement.props.style[1]).toEqual({ color: "#ffa500" });
+});
+
+test("Component maps style onto a same-named prop when nativeStyleMapping uses the `true` shorthand", () => {
+  registerCSS(`.fill { background-color: red; }`);
+
+  const mapping: StyledConfiguration<typeof RNView> = {
+    className: {
+      target: "style",
+      nativeStyleMapping: {
+        backgroundColor: true,
+      },
+    },
+  };
+
+  const Component = copyComponentProperties(
+    RNView,
+    (props: StyledProps<ViewProps, typeof mapping>) => {
+      return useCssElement(RNView, props, mapping);
+    },
+  );
+
+  const result = render(<Component testID={testID} className="fill" />);
+
+  const renderedProps = result.getByTestId(testID).props;
+
+  expect(renderedProps.backgroundColor).toBe("#f00");
+  expect(renderedProps.style?.backgroundColor).toBeUndefined();
 });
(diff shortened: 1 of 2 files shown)


danstepanov:

Thanks for this, the bug was real when you opened it, but #323 landed the same fix in June by normalizing the boolean shorthand in useNativeCss, with regression tests covering the TextInput case, and this branch now conflicts with main. I think this can close as superseded. Appreciate the clean test either way.

PR #379, opened 2026-07-17
fix: emit React Native border-start/end props for border-inline styles

### LINKED ISSUE #378: border-s-* and border-e-* compile to style props React Native does not support
Split out from nativewind/nativewind#1737, which was reported against v4 with a narrower symptom. The v5 version is broader.

All logical border utilities are silent no-ops on native. The compiler maps `border-inline-start-color` to `borderInlineStartColor`, and `border-s-1` produces `borderInlineStartWidth` and `borderInlineStartStyle`. None of these props exist in React Native. RN 0.81 ships logical inline props only for margin, padding, and inset. For borders it has `borderStartColor`, `borderEndColor`, `borderStartWidth`, and `borderEndWidth`, which are RTL aware. Unknown style keys get dropped by the view config filter, so nothing renders and nothing warns.

Verified by compiling `.x { border-inline-start-color: red; }` through the current compiler, which emits `borderInlineStartColor`, and by checking RN 0.81 view configs for the supported prop names. There is no remapping in the runtime either.

Fix should go in the declaration parsers in `src/compiler/declarations.ts` (the `border-inline-*` entries around lines 120 to 131): emit `border-start-color`, `border-end-color`, `border-start-width`, and `border-end-width` so they camelCase to the real RN logical props. Border style has no per-side RN equivalent, so the inline style declarations should fold into `borderStyle` or be dropped.

Note the existing tests in `src/__tests__/vendor/tailwind/borders.test.tsx` currently assert the invalid names (`border-s-1` expects `borderInlineStartWidth: 1`), so they pass while the behavior is broken. Expectations need to change with the fix.

The v4 issue is narrower and separate: only var() based custom colors fail there, because `unparsedPropertyMapping` in react-native-css-interop lacks the border inline entries that the parsed path has. That fix stays in the v4 branch.

danstepanov:
Closes #378, follow up to nativewind/nativewind#1737.

## What

border-inline-start-color, border-inline-end-color, and the width and shorthand variants compiled to props like borderInlineStartColor that React Native does not have, so every border-s-* and border-e-* utility (and border-x-*, which Tailwind v4 also emits as border-inline-*) was a silent no-op on native.

- Renames the inline logical border color and width properties to RN's RTL aware border-start-* and border-end-* props, in both the parsed path and the unparsed path that var() based colors take (the original nativewind/nativewind#1737 report).
- Fixes the border-inline, border-inline-start, and border-inline-end shorthands to emit those props, and fixes border-inline-width which previously only read the start value.
- Drops per-side border styles since RN only has a uniform borderStyle. solid is dropped silently because it matches the native default; anything else records a compiler warning.
- Dev builds now print compiler warnings from the metro transformer, so declarations without a native equivalent are visible while building instead of being collected and discarded.

## Tests

- New src/__tests__/compiler/logical-borders.test.ts covering colors (literal and var()), widths, shorthands, and the style drop and warning behavior.
- Updated borders.test.tsx expectations, which previously locked in the invalid borderInline* prop names.
- Full suite, typecheck, and lint pass locally. The three babel suites fail identically on clean main (missing --experimental-vm-modules in this environment) and are unrelated.

### DIFF
--- a/src/__tests__/compiler/logical-borders.test.ts
+++ b/src/__tests__/compiler/logical-borders.test.ts
@@ -1,0 +1,118 @@
+import { compile } from "react-native-css/compiler";
+
+const getRule = (css: string) => {
+  const compiled = compile(`.my-class { ${css} }`);
+  return {
+    rule: compiled.stylesheet().s?.find((rule) => rule[0] === "my-class")?.[1],
+    warnings: compiled.warnings(),
+  };
+};
+
+describe("logical border colors", () => {
+  test("border-inline-start-color", () => {
+    expect(getRule("border-inline-start-color: red;").rule).toStrictEqual([
+      { s: [1, 1], d: [{ borderStartColor: "#f00" }] },
+    ]);
+  });
+
+  test("border-inline-end-color", () => {
+    expect(getRule("border-inline-end-color: red;").rule).toStrictEqual([
+      { s: [1, 1], d: [{ borderEndColor: "#f00" }] },
+    ]);
+  });
+
+  test("border-inline-color", () => {
+    expect(getRule("border-inline-color: red;").rule).toStrictEqual([
+      { s: [1, 1], d: [{ borderStartColor: "#f00", borderEndColor: "#f00" }] },
+    ]);
+  });
+
+  // nativewind/nativewind#1737: colors that resolve at runtime take the
+  // unparsed path, which previously skipped the border-inline-* renames
+  test("border-inline-start-color with var()", () => {
+    expect(
+      getRule("border-inline-start-color: hsl(var(--primary));").rule,
+    ).toStrictEqual([
+      {
+        s: [1, 1],
+        d: [[[{}, "hsl", [{}, "var", "primary", 1]], "borderStartColor", 1]],
+        dv: 1,
+      },
+    ]);
+  });
+});
+
+describe("logical border widths", () => {
+  test("border-inline-start-width", () => {
+    expect(getRule("border-inline-start-width: 2px;").rule).toStrictEqual([
+      { s: [1, 1], d: [{ borderStartWidth: 2 }] },
+    ]);
+  });
+
+  test("border-inline-end-width", () => {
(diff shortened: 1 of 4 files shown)


YevheniiKotyrlo (commented):
Confirming from production (Expo SDK 55 · RN 0.83.6 · Tailwind v4) — `border-s-*` / `border-e-*` were silently dropped on native before this. Thanks for fixing it at the parser layer.

I compiled the full logical-border matrix through `fix-logical-border-props`. Longhands and literal shorthands are handled. One gap remains: **`var()`-valued `border-inline` shorthands still emit `borderInline*`.**

| declaration | this branch emits |
| --- | --- |
| `border-inline-start-color: hsl(var(--p))` — longhand | ✅ `borderStartColor` |
| `border-inline-color: red blue` — literal shorthand | ✅ `borderStartColor` + `borderEndColor` |
| `border-inline-color: hsl(var(--p))` — **var shorthand** | ❌ `borderInlineColor` |
| `border-inline-width: var(--w)` — **var shorthand** | ❌ `borderInlineWidth` |

A `var()` routes the shorthand to `parseUnparsedDeclaration`, where `propertyRename` maps only the longhands and `parseBorderInline*` never runs — so `border-inline-color` / `-width` fall through unrenamed. Not hypothetical: Tailwind v4's `border-x-<token>` emits exactly `border-inline-color: var(...)` (v4 theme colors are CSS vars), so a themed `border-x-primary` is still a native no-op after this PR.

`propertyRename` can't cover it — a shorthand needs a 1→2 expansion, not a 1→1 rename — so the fix goes in the same `parseUnparsedDeclaration`, mirroring your `parseBorderInline*`. I've opened it as #393, stacked on this branch so the diff is just the delta (+2 cases, `logical-borders` at 15/15). Merge it in, cherry-pick the commit, or tell me to retarget — whichever's easiest.

Out of scope by design: the full `border-inline: 2px solid var()` form (still emits `borderInline`; needs runtime shorthand splitting, and Tailwind never emits it) — a small follow-up if you want it.

Issue #380, opened 2026-07-17
babel import-plugin still rewrites react-native-web internals, causing a circular-import crash on web (incomplete fix of #196 / #202)


yucedemirayak:
### Before submitting a new issue

- [x] I tested using the latest version of the library, as the bug might be already fixed.
- [x] I tested using a [supported version](https://github.com/reactwg/react-native-releases/blob/main/docs/support.md) of react native.
- [x] I checked for possible duplicate issues, with possible answers.

### Bug summary

This is a follow-up to #196 — the fix in #202 is incomplete, so the react-native-web circular-import crash still reproduces on web with `react-native-css@3.0.7`.

On web, the app crashes at bundle load time:

```
Web  ERROR  [TypeError: Cannot read properties of undefined (reading 'default')]

Code: index.js
 35 | export { default as Button } from './exports/Button';
 36 | export { default as CheckBox } from './exports/CheckBox';
> 37 | export { default as FlatList } from './exports/FlatList';

Call Stack
  <global> (react-native-web/dist/index.js:37)
  <global> (react-native-css/dist/commonjs/components/FlatList.js:17)
  <global> (react-native-web/dist/vendor/react-native/Animated/components/AnimatedFlatList.js:12)
  <global> (react-native-web/dist/vendor/react-native/Animated/Animated.js:13)
```

(In a production bundle this surfaces as `can't access property "default", _exportsFlatList2 is undefined`.)

**Root cause**

There are two layers that rewrite `react-native(-web)` imports to `react-native-css/components/*`: the metro `webResolver` (`src/metro/resolver.ts`) and the babel `import-plugin` (`src/babel/import-plugin.ts`). #202 only touched the resolver, and only by excluding *resolution targets* inside `react-native-web/dist/vendor` (plus `VirtualizedList`). Neither layer excludes *origins* inside react-native-web, and the babel plugin was not changed at all.

The babel plugin's `parseReactNativeWebSource` resolves **relative** import specifiers against the current filename and rewrites anything that lands inside `react-native-web/dist`. So react-native-web's *own internal* imports get rewritten, which creates the cycle:

1. Something touches `Animated` → `react-native-web/dist/vendor/react-native/Animated/Animated.js` loads `AnimatedFlatList.js`.
[...]

PR #406, opened 2026-08-06
feat(vite): add a Vite resolver so className works outside Metro

### LINKED ISSUE #405: className is silently dropped under Vite: no bundler integration exists outside Metro
### Summary

Under Vite there is nothing that performs the `react-native` → `react-native-css/components` swap, so every `className` is dropped. The page still renders — same DOM, same text, no error, no warning — it is just unstyled. That makes it easy to ship without noticing.

Opening this before a PR, per CONTRIBUTING's guidance to discuss API changes first. I have a working implementation and would be glad to submit it, adjust the approach, or drop it if you would rather solve this differently.

**Reproduction:** https://github.com/p-larson/nativewind-vite-storybook-repro

Two apps, byte-identical except for one Vite plugin, plus a script that asserts computed styles rather than "did it render":

```
[broken]  className in DOM: false   computed size: null x null
[fixed]   className in DOM: true    computed size: 320px x 160px
```

| broken | fixed |
|---|---|
| unstyled | styled |

### Why it happens

NativeWind 4 exported a JSX runtime, so the documented [Other Bundlers](https://nativewind.dev/docs/guides/other-bundlers) recipe worked anywhere. Its third requirement is no longer satisfiable on v5 — `nativewind@5.0.0-preview.4` exports only `.`, `./babel`, `./metro`, `./types`, `./theme`, and setting `jsxImportSource: "nativewind"` fails the build:

```
Error: Missing "./jsx-dev-runtime" specifier in "nativewind" package
```

The work moved here, into a Metro-only resolver. I understand from the README that Metro is the officially supported bundler and that contributions for others are welcome — this is that contribution.

### Proposed approach

Add a `react-native-css/vite` export applying the mapping `nativeResolver` already uses:

```ts
if (moduleName === "react-native") {
  return resolver(context, `react-native-css/components`, platform);
}
```

`react-native-css/components` is already a barrel re-exporting `react-native` with the wrappers layered on top, so no new module is introduced and the component list cannot drift.

I first tried porting `webResolver` (rewriting `react-native-web/dist/exports/<X>/index.js` → `components/<X>`) and abandoned it:

1. **Rollup resolvers do not run during dependency pre-bundling.** Making the rewrite fire requires excluding `react-native-web` from `optimizeDeps`, which exposes its CommonJS dependencies to the browser as raw CJS — `inline-style-prefixer`, `fbjs`, `styleq`, and others, each needing its own `optimizeDeps.include` entry, down to deep paths like `inline-style-prefixer/lib/plugins/crossFade.js`.

2. **It rewrites `react-native-web`'s own internal imports**, producing exactly the initialization cycle in #380.

Following `nativeResolver` avoids both — `react-native-web` stays pre-bundled and its internals are never touched.

Two details that are easy to get wrong:

- Imports originating inside `react-native-css` must be skipped (mirroring `isFromThisModule`). The barrel re-exports `react-native` and each wrapper uses its base component at module scope, so redirecting them recreates the #380 cycle.
- Both `react-native` and `react-native-web` must be matched, because `vite-plugin-rnw` applies that alias before user plugins run. Matching only `react-native` silently no-ops.

The mapping needs registering twice: as a Vite plugin for source files, and as an esbuild plugin in `optimizeDeps.esbuildOptions.plugins` for pre-bundled dependencies, since esbuild does not run Rollup resolvers.

### Status

Implemented on [`p-larson/react-native-css@feat/vite-resolver`](https://github.com/p-larson/react-native-css/tree/feat/vite-resolver): `yarn lint`, `yarn typecheck`, and the full `yarn test` suite (1056 passing) are green, with 5 new resolver tests. `vite` is added as an *optional* peer dependency so Metro-only users are unaffected. I also packed the branch and ran the reproduction against the real `react-native-css/vite` export end-to-end.

Happy to open the PR if the approach looks reasonable.

### Environment

| Package | Version |
|---|---|
| `react-native-css` | 3.0.7 |
| `nativewind` | 5.0.0-preview.4 |
| `react-native-web` | 0.21 |
| `vite` | 7.2 |
| `@storybook/react-native-web-vite` | 10.2 |

p-larson:
Closes #405 — opened there first per CONTRIBUTING's guidance on API changes. Happy to rework or close this if you would rather take a different approach.

## Problem

Under Vite nothing performs the `react-native` → `react-native-css/components` swap, so every `className` is dropped. The page still renders — same DOM, same text, no error and no warning — it is just unstyled.

NativeWind 4 exported a JSX runtime, so the documented [Other Bundlers](https://nativewind.dev/docs/guides/other-bundlers) recipe worked anywhere. On v5 its third requirement cannot be satisfied: `nativewind@5.0.0-preview.4` exports only `.`, `./babel`, `./metro`, `./types`, `./theme`, and setting `jsxImportSource: "nativewind"` fails with `Missing "./jsx-dev-runtime" specifier in "nativewind" package`.

## Approach

Adds a `react-native-css/vite` export applying the mapping `nativeResolver` already uses:

```ts
if (moduleName === "react-native") {
  return resolver(context, `react-native-css/components`, platform);
}
```

`react-native-css/components` is already a barrel re-exporting `react-native` with the wrappers layered on top, so no new module is introduced and the component list cannot drift as wrappers are added.

### Why not port `webResolver`

Rewriting `react-native-web/dist/exports/<X>/index.js` → `components/<X>` was tried first and abandoned:

1. **Rollup resolvers do not run during dependency pre-bundling.** Making the rewrite fire requires excluding `react-native-web` from `optimizeDeps`, which exposes its CommonJS dependencies to the browser as raw CJS — `inline-style-prefixer`, `fbjs`, `styleq` and others, each needing its own `optimizeDeps.include` entry, down to deep paths like `inline-style-prefixer/lib/plugins/crossFade.js`.
2. **It rewrites `react-native-web`'s own internal imports**, producing the initialization cycle reported in #380.

Following `nativeResolver` avoids both: `react-native-web` stays pre-bundled and its internals are never touched.

### Implementation notes

- **Imports originating inside `react-native-css` are skipped**, mirroring `isFromThisModule`. The barrel re-exports `react-native` and each wrapper uses its base component at module scope (`copyComponentProperties(RNView, …)` runs during evaluation), so redirecting them recreates the #380 cycle.
- **Both `react-native` and `react-native-web` are matched**, because `vite-plugin-rnw` applies that alias before user plugins run. Matching only `react-native` silently no-ops.
- **The mapping is registered twice** — as a Vite plugin for source files, and as an esbuild plugin in `optimizeDeps.esbuildOptions.plugins` for pre-bundled dependencies, since esbuild does not run Rollup resolvers.
- **`vite` is an optional peer dependency**, so Metro-only users are unaffected.

## Testing

- 5 new tests in `src/__tests__/vite/resolver.test.ts` covering both specifiers, the `isFromThisModule` guard, unrelated specifiers, and the optimizeDeps registration.
- `yarn lint`, `yarn typecheck`, and `yarn test` (1056 passing, 56 suites) all green.
- Packed this branch and ran it end-to-end through the real `react-native-css/vite` export against a reproduction repo: https://github.com/p-larson/nativewind-vite-storybook-repro

That repo has two apps, byte-identical except for this plugin, and asserts computed styles rather than "did it render":

```
[broken]  className in DOM: false   computed size: null x null      PASS (expected unstyled)
[fixed]   className in DOM: true    computed size: 320px x 160px    PASS (expected styled)
```

| broken | fixed |
|---|---|
| unstyled | styled |

## Docs

README's "Other bundlers" section is replaced with a "Vite based projects" section covering plain Vite and Storybook's `react-native-web-vite`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,41 @@ Follow the Expo instructions, but replace the `expo` package with `@expo/metro-c
 + import { getDefaultConfig } from "@expo/metro-config";
 ```
 
+### Vite based projects
+
+Add the `reactNativeCSS` plugin to your Vite configuration:
+
+```ts
+import { defineConfig } from "vite";
+import { reactNativeCSS } from "react-native-css/vite";
+
+export default defineConfig({
+  plugins: [reactNativeCSS()],
+});
+```
+
+For Storybook's `react-native-web-vite` framework, add it in `viteFinal`:
+
+```ts
+import { mergeConfig } from "vite";
+import { reactNativeCSS } from "react-native-css/vite";
+
+const config = {
+  framework: "@storybook/react-native-web-vite",
+  viteFinal: (config) =>
+    mergeConfig(config, { plugins: [reactNativeCSS()] }),
+};
+
+export default config;
+```
+
+Vite does not process CSS through the Metro transformer, so import your
+Tailwind entry stylesheet directly (for example in `.storybook/preview.ts`)
+and let Vite's PostCSS pipeline handle it.
+
 ### Other bundlers
 
-`react-native-css` officially only supports Metro as the bundler, but we welcome community contributions to support other bundlers like Webpack, Vite or Turbopack.
+`react-native-css` officially supports Metro and Vite, but we welcome community contributions to support other bundlers like Webpack or Turbopack.
 
 More documentation coming soon.
 
--- a/package.json
+++ b/package.json
@@ -94,6 +94,17 @@
         "default": "./dist/commonjs/metro/index.js"
       }
     },
+    "./vite": {
+      "source": "./src/vite/index.ts",
+      "import": {
+        "types": "./dist/typescript/module/src/vite/index.d.ts",
+        "default": "./dist/module/vite/index.js"
+      },
+      "require": {
+        "types": "./dist/typescript/commonjs/src/vite/index.d.ts",
+        "default": "./dist/commonjs/vite/index.js"
+      }
+    },
     "./native": {
       "source": "./src/native/index.ts",
       "import": {
@@ -207,7 +218,13 @@
     "@expo/metro-config": ">=54",
     "lightningcss": ">=1.27.0",
     "react": ">=19",
-    "react-native": ">=0.81"
+    "react-native": ">=0.81",
+    "vite": ">=5"
+  },
+  "peerDependenciesMeta": {
+    "vite": {
+      "optional": true
+    }
   },
   "devDependencies": {
     "@babel/core": "^7.28.0",
@@ -250,7 +267,8 @@
     "tailwindcss": "^4.1.12",
     "tailwindcss-safe-area": "^1.1.0",
     "typescript": "^5.9.2",
-    "typescript-eslint": "^8.40.0"
+    "typescript-eslint": "^8.40.0",
+    "vite": "^7.2.2"
   },
   "react-native-builder-bob": {
     "source": "src",
--- a/src/__tests__/vite/resolver.test.ts
+++ b/src/__tests__/vite/resolver.test.ts
@@ -1,4 +1,4 @@
-import { sep } from "node:path";
+import { resolve as resolvePath } from "node:path";
 
 import { reactNativeCSS } from "../../vite";
 
@@ -20,8 +20,11 @@ function setup() {
   return { plugin, resolve, call };
 }
 
-const rncFile = `${sep}node_modules${sep}react-native-css${sep}dist${sep}components${sep}View.js`;
-const appFile = `${sep}app${sep}src${sep}Button.tsx`;
+/** Inside this package, so `isFromThisModule` must skip it. */
+const ownFile = resolvePath(
+  "/app/node_modules/react-native-css/dist/module/components/View.js",
+);
+const appFile = resolvePath("/app/src/Button.tsx");
 
 describe("vite resolver", () => {
   it("redirects react-native to react-native-css/components", async () => {
@@ -48,12 +51,21 @@ describe("vite resolver", () => {
     );
   });
 
-  it("does not redirect imports from react-native-css itself", async () => {
+  it("does not redirect imports from this module", async () => {
     const { resolve, call } = setup();
 
     // The components barrel re-exports react-native and each wrapper uses its
     // base component at module scope, so redirecting these would cycle.
-    await expect(call("react-native", rncFile)).resolves.toBeNull();
+    await expect(call("react-native", ownFile)).resolves.toBeNull();
+    expect(resolve).not.toHaveBeenCalled();
+  });
+
+  it("strips Vite's query suffix before checking the importer", async () => {
+    const { resolve, call } = setup();
+
+    await expect(
+      call("react-native", `${ownFile}?v=abc123`),
+    ).resolves.toBeNull();
     expect(resolve).not.toHaveBeenCalled();
   });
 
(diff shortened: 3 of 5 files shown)


p-larson:
Following up on the obvious question — **why a resolver rather than the existing `react-native-css/babel` plugin**, which already does this rewrite bundler-agnostically. I evaluated it rather than assuming, and added [`apps/babel`](https://github.com/p-larson/nativewind-vite-storybook-repro/tree/main/apps/babel) to the reproduction.

It does work for app source. It is the weaker layer for one structural reason, plus two packaging bugs that are worth fixing regardless of this PR.

### It does not reach code inside dependencies

`@vitejs/plugin-react` only runs babel over files matching its include filter, which excludes `node_modules`. Adding a published package that forwards `className` and measuring both approaches on the same story:

| | app source | code inside a dependency |
|---|---|---|
| Resolver (this PR) | `320px x 160px` | `160px x 80px` |
| `babel/import-plugin` | `320px x 160px` | class never reaches the DOM |

Widening the filter to cover `node_modules` is possible, but that is exactly what re-introduces #380 — once babel rewrites react-native-web's own internal imports, the cycle appears. The resolver never rewrites anything inside react-native-web, so it avoids that by construction rather than by exclusion list.

Resolution also happens once per specifier instead of an AST pass per module.

### Two packaging bugs found while testing

These are independent of this PR — happy to split them into their own issue, or fix them here if you prefer.

**1. The ESM build of the babel preset calls `require`.** Vite resolves the `import` condition to `dist/module/babel/index.js`, which fails immediately:

```
Vite Pre-transform error: [BABEL] require is not defined
(While processing: react-native-css/dist/module/babel/index.js)
```

Source is `src/babel/index.ts`:

```ts
export default function () {
  return {
    plugins: [
      require("./import-plugin").default,
      "react-native-worklets/plugin",
    ],
  };
}
```

Under Metro this resolves to the CommonJS build so it never surfaces, but the ESM build is unusable by any ESM-based babel config loader.

**2. `import-plugin` has no public export.** The only babel entry is the preset, which also hard-requires `react-native-worklets/plugin` — not needed on web. Loading just the import-plugin means going through `require.resolve("react-native-css/package.json")` and joining a `dist/commonjs/...` path by hand, since the exports map rejects the deep path with `ERR_PACKAGE_PATH_NOT_EXPORTED`.

If you would rather solve Vite through babel than a resolver, fixing those two would be the prerequisite — though the dependency-coverage gap above would remain.

p-larson:
One refinement pushed, and it surfaces a latent issue in the existing code worth flagging separately.

I initially copied `isFromThisModule` verbatim from `src/metro/resolver.ts`:

```ts
const thisModuleDist = resolve(__dirname, "../../../dist");
```

That depth is correct for the built output (`dist/commonjs/metro/resolver.js` → package root), but not when the module is loaded from `src` via the `source` export condition:

```
built  dist/commonjs/metro  -> thisModuleDist = <pkg>/dist
source src/metro            -> thisModuleDist = /tmp/dist      # outside the package
```

So from `src` the guard silently never matches. My tests caught it because jest runs against `src`. Since Vite resolves through whichever condition the consumer configured, the Vite plugin now anchors on the package directory instead, which holds for both layouts, and there is a test for Vite's `?v=` query suffixes.

The same appears to apply to `src/metro/resolver.ts` and `src/babel/import-plugin.ts` under the `source` condition. I have deliberately left those alone to keep this PR to one change — happy to open a separate issue or PR if it is worth fixing.

Full suite still green: 1057 passing, 56 suites, plus `yarn lint` and `yarn typecheck`.

PR #410, opened 2026-08-14
fix(native): share the :root variable registries across a dual-package split

YevheniiKotyrlo:
## Problem

When a bundle contains **both** builds — `dist/module` for ESM importers and `dist/commonjs` for CommonJS ones — the two copies get **separate `:root` variable registries**. The compiled stylesheet is injected exactly once, into whichever copy the app's `.css` module resolved. Every component resolving through the *other* copy then looks up `var(--anything)` in an empty registry, gets `undefined`, and silently renders with no value.

For a `className` colour that means the element falls back to React Native's default text colour — black on Android, invisible on any dark surface. Nothing errors; the class is found (the style registry IS shared), only its *value* resolves to nothing.

This reproduces in a stock Expo SDK 57 app for any component shipping compiled CommonJS. `expo-router`'s `<Link>` is the case that surfaced it.

### Measured

Android, dark theme, `text-danger-foreground` (`#fb2c36`), sampled from the rendered pixels rather than by eye:

```
before:  painted=rgb(0,0,0)      bg=rgb(22,22,25)   contrast=39.9    ← effectively invisible
after:   painted=rgb(251,44,54)  bg=rgb(22,22,25)   contrast=231.9
```

Both builds are genuinely in the graph — resolving each module's dependency map to its path in a real Metro dev bundle:

```
styles/global.css        -> dist/module/native-internal/index.js      ← stylesheet injected HERE
first-party .tsx source  -> dist/module/components/index.cjs          ← same copy, works
expo-router BaseExpoRouterLink.js -> dist/commonjs/components/index.cjs ← OTHER copy, empty registry
```

## Root cause

`src/native-internal/root.ts` creates its registries at module scope, so two copies of the module means two independent registries. `exports` splits `import`/`require` across builds and Metro resolves that condition per *requesting* module.

Two sibling modules already guard against exactly this — `native-internal/style-collection.ts` and `native-internal/variables.ts` both use `globalThis.… ??=`. `root.ts` was missed.

## Why the one-line `??=` is a regression

`root.ts` does not only create the registries, it **seeds** them (`__rn-css-rem`, and the platform `__rn-css-color`). `__rn-css-rem` is also set by the injected stylesheet — a project pinning `:root { font-size: 16px }` injects `[[16]]`. So a bare `globalThis.x ??= rootVariableFamily()` leaves the seeds running unconditionally, and a second copy initialising *after* the inject re-runs `set([[14]])` and clobbers 16 back to 14 — silently rescaling every rem-derived utility to 87.5%.

Creation and seeding therefore have to be one atomic, once-only step, which is what this does.

## Tests

Three added, each mutation-proven (I broke the fix and watched the assertion go red before trusting it). Suite, typecheck and lint sit at the pristine-`main` baseline.

### DIFF
--- a/src/__tests__/native/root-variable-registry.test.ts
+++ b/src/__tests__/native/root-variable-registry.test.ts
@@ -1,0 +1,47 @@
+import {
+  resolveRootVariableRegistries,
+  rootVariables,
+  universalVariables,
+} from "../../native-internal/root";
+
+/**
+ * The `:root` registries are global, and created + seeded exactly once.
+ *
+ * The package's `exports` map splits `import` and `require` onto different
+ * builds, and Metro resolves that condition per requesting module — so a
+ * compiled-CommonJS dependency and first-party source bind different copies of
+ * `native-internal/root`. Module-scope registries made that two independent
+ * stores: a `:root` variable injected into one copy was invisible to the
+ * other, and the value silently fell back to its seed.
+ *
+ * `resolveRootVariableRegistries` is what a second copy of the module runs, so
+ * calling it directly reproduces the second copy without needing the module
+ * itself to be re-evaluated.
+ */
+test("the module's exports are the registries published on globalThis", () => {
+  const registries = globalThis.__react_native_css_root_variable_registries;
+
+  expect(registries).toBeDefined();
+  expect(rootVariables).toBe(registries?.root);
+  expect(universalVariables).toBe(registries?.universal);
+});
+
+test("a second copy resolves the same registries", () => {
+  // Captured BEFORE the call: comparing against the global afterwards passes
+  // even when the resolver replaces it, which is no assertion at all.
+  const before = globalThis.__react_native_css_root_variable_registries;
+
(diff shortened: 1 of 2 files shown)


YevheniiKotyrlo:
### Where the device evidence for this one is

This PR does carry a device measurement — it is in the body rather than here, under **Measured**, because it is four lines of sampled pixels rather than an image pair.

```
before:  painted=rgb(0,0,0)      bg=rgb(22,22,25)   contrast=39.9    ← effectively invisible
after:   painted=rgb(251,44,54)  bg=rgb(22,22,25)   contrast=231.9
```

Android, dark theme, `text-danger-foreground`, read off the rendered pixels rather than judged by eye. A screenshot pair would show the same thing less precisely: the "before" frame is a dark surface with text on it you cannot see, which photographs as a dark surface.

Gate results at this branch's head, on Windows:

```
tsc --noEmit   0 errors
eslint         0 problems
jest           1050 passed, 3 failed, 21 skipped
```

The three are the Windows-only `babel-plugin-tester` cases this branch does not touch — #390 is the branch that fixes them.

PR #420, opened 2026-08-15
fix(compiler): make a light-dark() extra rule a rule in its own right

YevheniiKotyrlo:
## Problem

`light-dark()` compiles to two rules: the current rule carrying the light branch, and an extra copy of it under `prefers-color-scheme: dark` carrying the dark branch. The extra rule is never built or applied as a rule in its own right — it is seeded from the current rule mid-parse (`extendRule`), then patched onto a clone of the rule that has already been applied to the selector (`createRuleFromPartial`). Five consequences, all silent, all in dark mode.

### 1. A second `light-dark()` declaration re-asserts the first one's light value

`parseUnresolvedColor` opens its extra rule with `extendRule({ m: [...] })`, which is `{ ...this.rule, ...rule }` — a snapshot of everything already written to the rule. The second declaration's dark rule therefore carries the first declaration's **light** descriptor, and being applied last, it wins.

Two `light-dark()` colours on one rule is ordinary in a themed stylesheet, and a `var()` anywhere inside either branch is what keeps the colour unresolved and routes it through this path:

```css
:root { --o: 1; }
.parent {
  color: light-dark(hsl(0 100% 50% / var(--o)), hsl(240 100% 50% / var(--o)));
  background-color: light-dark(hsl(120 100% 25% / var(--o)), hsl(60 100% 50% / var(--o)));
}
```

Rendered through `@testing-library/react-native` with `colorScheme.set("dark")`:

| | `main` (f70c402) | this branch |
| --- | --- | --- |
| light | `color: hsl(0, 100, 50, 1)`, `backgroundColor: hsl(120, 100, 25, 1)` | same |
| dark | `color: `**`hsl(0, 100, 50, 1)`**, `backgroundColor: hsl(60, 100, 50, 1)` | `color: `**`hsl(240, 100, 50, 1)`**, `backgroundColor: hsl(60, 100, 50, 1)` |

The background colour switches and the text colour does not. Not because `color`'s dark rule is wrong — it is correct and it is applied — but because the background's dark rule then re-applies the light red over it.

### 2. The dark rule publishes the light `--__rn-css-color`

`createRuleFromPartial` copies `m` and `d` from the extra rule and nothing else, so `v` — the variables the rule publishes to its subtree — stays whatever the base rule's clone held. `--__rn-css-color` is the channel `currentcolor` and inherited colour resolve through, so on `main` `.a { color: light-dark(red, blue) }` **renders `#00f` and tells every descendant `#f00`**.

### 3. `dv` does not survive either, so a dark-only `var()` is not flagged for late resolution

`.c { background-color: light-dark(red, var(--d)); }` — only the dark branch reads a variable, so `dv` (resolve this rule's declarations late) is the extra rule's own requirement; the rule it copies has no delayed declaration to inherit the flag from. On `main` the dark rule comes out `dv=undefined`; on this branch `dv=1`.

### 4. The LAST extra rule on a rule republishes the base rule's variable

Fixing 2 by replacing `v` from the extra rule leaves a fallback: when the extra rule publishes nothing, the merged rule keeps the base rule's `v`. That is invisible for a single extra rule and wrong for the last of several. `background-color` publishes nothing, so its extra rule — opened after `color`'s, and applied after it — carries a copy of the base rule's **light** colour over the dark one `color`'s extra rule had already published.

```css
.parent { color: light-dark(red, blue); background-color: light-dark(green, yellow); }
.child  { background-color: currentcolor; }
```

| | `main` | this branch |
| --- | --- | --- |
| dark, parent | `color: #00f`, `backgroundColor: #ff0` | same |
| dark, child | `backgroundColor: `**`#f00`** | `backgroundColor: `**`#00f`** |

### 5. Inside a pseudo-element, the dark half lands on the element

`::selection` and `::placeholder` map their `color` onto a prop of the host component. That mapping is applied to the rule on its way to the selector — and the extra rule is composed **after** it, replacing `d` wholesale with a partial that never went through it. So the light half is scoped and the dark half is a plain `color` on the element.

```
.a::selection { color: light-dark(red, blue); }
```

| | `main` | this branch |
| --- | --- | --- |
| light | `d=[{}, ["#f00",["selectionColor"]]]` | same |
| dark | `d=`**`[{"color":"#00f"}]`** | `d=[{}, `**`["#00f",["selectionColor"]]`**`]` |

Rendered on a `TextInput`, `main` in dark mode gives `selectionColor: "#f00"` — still the light one, because the dark rule never produced a `selectionColor` — **and** `style: { color: "#00f" }`, which recolours the input's own text. `::placeholder` behaves identically with `placeholderTextColor`.

This is the defect #411 exists to close, on a path its field policy never sees: #411 scopes the rule, and the extra rule is composed after the scoping ran.

## Root cause

One site, two directions. The extra rule is composed onto the rule that has already been applied to the selector, by patching a clone of it — so it **inherits content it must not** (1, 2, 3, 4) and **skips shaping it must have** (5).

Two smaller ones sit under that:

- `builder.descriptorProperty` is a single property, and `addUnnamedDescriptor` — the seam a `light-dark()` dark branch is written through — reaches only that one. A `color` declaration writes twice (the style property, and the variable it publishes), so one of the two was always going to be missed.
- `parseFontColorDeclaration` parses its colour twice, once per write, and `parseColor` is not pure — a `light-dark()` value opens an extra rule. So a `color: light-dark()` emitted two identical dark rules.

## Fix

**`openExtraRule(condition)`** creates the extra rule **empty** and hands it back for the caller to write descriptors into through the same `addDescriptor` seams as the current rule. Both call sites in `parseColor` / `parseUnresolvedColor` use it; `extendRule` and `addExtraRule` go away.

**`mergeExtraRule`** builds from the extra rule rather than patching the applied one. The rule it was opened on supplies the SELECTOR — its specificity, its container query, and the media conditions the extra one is added to. The extra rule supplies the CONTENT, in full, and inherits none of it: the rule it was opened on matches under the extra condition too, so anything the extra rule leaves out still arrives from there, while restating it makes the extra rule a second place that value is written.

Pseudo classes and attribute queries are not in that list because they are never on the rule the merge copies from. `applyRuleToSelectors` reads them off the SELECTOR and writes them onto every rule it applies, the merged one included, so a copy in the merge would be a second source for a value that already arrives one step later. A tripwire on `rule.p` / `rule.aq` inside the merge fired on no test in the suite and on none of 20 targeted probe cases; the same tripwire on `rule.m` / `rule.cq` fired on five, so the instrument was doing its job.

**`applyRuleToSelectors`** puts every rule a selector receives through one pipeline. The extra rules are merged up front, then each source rule is cloned, scoped to the pseudo-element, and given the selector's specificity and queries by the same code. The parent classes a container query names describe the selector rather than the rule, so they are still registered once however many rules it receives.

**`descriptorProperty` becomes `descriptorProperties: readonly string[]`**, and `parseFontColorDeclaration` / `parseUnparsedDeclaration` name both `color` and `--__rn-css-color` on it. `StylesheetBuilder` is not re-exported from `react-native-css/compiler`, so that rename is internal.

**`parseFontColorDeclaration` parses its colour once** and uses the value for both writes.

## Which plane

**Compiler** (`src/compiler/declarations.ts`, `src/compiler/stylesheet.ts`), and it reaches only the native runtime — the compiled rule set is what `src/native` consumes. Web needs no mirror: it serves the CSS to the browser, which implements `light-dark()` itself. There is no `light-dark()` handling anywhere under `src/web`.

## Tests

42 in two files. **24 fail on `main`**, 18 pass.

- `src/__tests__/compiler/light-dark.test.ts` — 28 tests, **17 fail on `main`**. What each dark rule is allowed to contain: its own declaration only, its own `dv`, its own published colour, nothing restated from the rule it copies, and the pseudo-element mapping its rule was given.
- `src/__tests__/native/light-dark.test.tsx` — 14 tests, **7 fail on `main`**. The rendered pairs from the tables above, a dark-branch `var()` asserted against the same variable read outside `light-dark()`, inheritance driven across `colorScheme.set`, and `::selection` / `::placeholder` on a `TextInput` in both schemes.

Six of the 28 pass on both refs, deliberately. `mergeExtraRule` states the selector context as an enumerated field list, so a field dropped from that list is a silent behaviour change rather than a compile error — those six hold the fields it decides. Each was measured by removing the behaviour and confirming which tests redden, over the whole suite:

| Removed | Reddens |
| --- | --- |
| the `cq` carry in `mergeExtraRule` | 1 — the dark rule of the `@container` case loses its container query, so it applies outside the container |
| the `m` carry in `mergeExtraRule` | 1 — the dark rule of the `@media` case loses `[">=","width",100]`, so it applies at any width |
| the once-per-selector container guard | 1 — a selector with three `light-dark()` declarations registers its parent container class 4× instead of 1× |
| the selector's pseudo class on every rule but the first | 1 |
| the selector's attribute query on every rule but the first | 1 |

Nothing else reddens in any of the five, and each of those five reddened **nothing at all** before these tests existed. Each assertion is stated over the light rule rather than over a literal, so what it measures is the carry and not the encoding.

Every light-mode case is stated beside its dark one and stays green throughout. It is not filler — the light half is the surface an over-broad fix breaks.

Full suite, typecheck and lint measured on the same machine and worktree layout, no new failures. `main` is 1048 passed / 3 failed (the two `src/__tests__/babel/*` suites, which fail identically at every ref on Windows); this branch is `Test Suites: 2 failed, 4 skipped, 55 passed, 57 of 61 total` / `Tests: 3 failed, 21 skipped, 1090 passed, 1114 total` — the whole +42 is these two files. `yarn typecheck` and `yarn lint` exit 0 on both.

## KNOWN LIMITS

None outstanding in `light-dark()` handling. Three things adjacent to it are deliberately left alone:

- **A shorthand emits one dark rule per longhand it expands to.** `parseColor` opens an extra rule as a side effect of parsing, so a caller that parses the same value more than once opens that many — which is defect 6 above, reached by a different caller. Measured identically on `main` (f70c402) and on this branch: `border-color` emits **4** identical dark rules, `border-inline-color` and `border-block-color` **2** each. Every other colour property probed emits 1: `background-color`, the four physical `border-*-color` longhands, the four flow-relative single-side longhands, `text-decoration-color`, `caret-color`, `outline-color`, `fill`, `stroke`. Both refs agree on every row, so none of it is a regression this PR introduces; the one property in that census this PR changes is `color`, which goes from **2** dark rules to **1**.
- **A pseudo-element rule still publishes `--__rn-css-color`.** `.a::selection { color: red }` scopes its declaration and still hands the element's subtree the colour, because the scoping rewrites `d` and leaves `v`. That is the same fault on a channel this PR does not touch, and #411's field policy is what closes it — for the base rule and, once both land, for the extra rule too, since they now go through one call site.
- **`::selection`'s mapping is `color` → `selectionColor`.** #411 argues that is the wrong pair and changes it to `background-color`. This PR only makes the dark half take whichever mapping the light half took.

**On the version bump.** No `BREAKING CHANGE:` footer, so the preset recommends `patch`. Defect 5 does remove behaviour — a dark-mode `color` that used to reach the element no longer does — but the light half never reached it either, so this makes the two halves of one declaration agree rather than changing a contract anyone could have relied on. Say the word if you would rather it were marked.

---

**Overlaps with open PRs.** Measured with `git merge-tree` against every open PR head, at this branch's head:

- **#391** (`fix/color-inherit`) — hard conflict in `src/compiler/declarations.ts`. Both touch the inherited-colour publish path; whichever lands second needs a real rebase, not a mechanical one.
- **#393** (`fix/border-inline-var-shorthand`) — hard conflict in `src/compiler/declarations.ts` and `src/compiler/stylesheet.ts`.
- **#411** (`fix/pseudo-element-declaration-leak`) — hard conflict in `src/compiler/stylesheet.ts`, **new with defect 5's fix** and unavoidable: #411 replaces the pseudo-element scoping call, and this PR changes which rules reach it. Both conflicts are inside `applyRuleToSelectors` and the composition is mechanical — #411's `scopeRuleToPseudoElement` and its warning bookkeeping go inside this PR's `for (const sourceRule of sourceRules)` loop, which is the desired result: #411's field policy then covers extra rules too, and its `warnedPseudoElements` set already dedupes per pseudo-element across the whole method, so the extra rules add no second warning.
- **#412**, **#413**, **#417**, **#389**, **#397**, **#421**, **#422**, **#423**, **#415**, **#410**, **#418**, **#416**, **#388**, **#392**, **#390** all auto-merge clean.

### DIFF
--- a/src/__tests__/compiler/light-dark.test.ts
+++ b/src/__tests__/compiler/light-dark.test.ts
@@ -1,0 +1,202 @@
+import { compile } from "react-native-css/compiler";
+import type {
+  CompilerOptions,
+  StyleDeclaration,
+  StyleDescriptor,
+  StyleRule,
+} from "react-native-css/compiler";
+
+/**
+ * `light-dark()` compiles to two rules: the current one, carrying the light
+ * branch, and an extra copy of it under `prefers-color-scheme: dark` carrying
+ * the dark branch. These helpers read the pair back out of a compiled
+ * stylesheet so a test can state what each half is allowed to contain.
+ */
+const rulesFor = (
+  css: string,
+  className: string,
+  options?: CompilerOptions,
+): StyleRule[] => {
+  const ruleSets = compile(css, options).stylesheet().s ?? [];
+  return ruleSets.flatMap(([name, rules]) => (name === className ? rules : []));
+};
+
+const isDarkRule = (rule: StyleRule): boolean =>
+  Boolean(
+    rule.m?.some(
+      (condition) =>
+        condition[0] === "=" &&
+        condition[1] === "prefers-color-scheme" &&
+        condition[2] === "dark",
+    ),
+  );
+
+const darkRules = (
+  css: string,
+  className: string,
+  options?: CompilerOptions,
+): StyleRule[] => rulesFor(css, className, options).filter(isDarkRule);
+
+const lightRule = (
+  css: string,
+  className: string,
+  options?: CompilerOptions,
+): StyleRule => {
+  const rule = rulesFor(css, className, options).find(
+    (candidate) => !isDarkRule(candidate),
+  );
+
+  if (!rule) {
+    throw new Error(`No unconditional rule for .${className}`);
+  }
+
+  return rule;
+};
+
+/**
+ * The style properties a rule sets. A declaration is either a static record of
+ * property/value pairs or a `[value, propertyPath]` tuple, and both forms
+ * appear in the same list.
+ */
+const declaredProperties = (rule: StyleRule): string[] => {
+  return (rule.d ?? []).flatMap((declaration: StyleDeclaration) => {
+    if (Array.isArray(declaration)) {
+      const propertyPath = declaration[1];
+      return typeof propertyPath === "string"
+        ? [propertyPath]
+        : [propertyPath.join(".")];
+    }
+
+    return Object.keys(declaration);
+  });
+};
+
+const variable = (
+  rule: StyleRule,
+  name: string,
+): StyleDescriptor | undefined => {
+  return rule.v?.find(([variableName]) => variableName === name)?.[1];
+};
+
+describe("an extra rule carries only its own declaration", () => {
+  /**
+   * A `var()` anywhere inside `light-dark()` keeps the colour unresolved, which
+   * is the path that opens the extra rule from the current rule rather than
+   * from an empty one. Two such declarations on one rule is the trigger: a
+   * colour and a background colour together is ordinary in a themed stylesheet.
+   */
+  const twoUnresolvedDeclarations = `
+.p1 {
+  color: light-dark(hsl(0 100% 50% / var(--a)), hsl(240 100% 50% / var(--a)));
+  background-color: light-dark(hsl(120 100% 25% / var(--a)), hsl(60 100% 50% / var(--a)));
+}`;
+
+  test("each declaration opens one dark rule", () => {
+    expect(darkRules(twoUnresolvedDeclarations, "p1")).toHaveLength(2);
+  });
+
+  test("a dark rule sets the properties of its own declaration, and no others", () => {
+    expect(
+      darkRules(twoUnresolvedDeclarations, "p1").map(declaredProperties),
+    ).toStrictEqual([["color"], ["backgroundColor"]]);
+  });
+
+  test("the second declaration's dark rule does not re-assert the first's light value", () => {
+    const [, backgroundDarkRule] = darkRules(twoUnresolvedDeclarations, "p1");
+
+    expect(backgroundDarkRule?.d).toStrictEqual([
+      [[{}, "hsl", [60, 100, 50, [{}, "var", "a", 1]]], "backgroundColor", 1],
+    ]);
+  });
+});
+
+describe("an extra rule carries its own flags", () => {
+  /**
+   * Only the dark branch reads a variable, so `dv` is the extra rule's own — the
+   * rule it copies has no delayed declaration to inherit the flag from.
+   */
+  const darkBranchVariable = `.p2 { background-color: light-dark(red, var(--d)); }`;
+
+  test("the light rule needs no delayed resolution", () => {
+    expect(lightRule(darkBranchVariable, "p2").dv).toBeUndefined();
+  });
+
+  test("the dark rule declares delayed resolution for its own variable", () => {
+    expect(darkRules(darkBranchVariable, "p2")).toHaveLength(1);
+    expect(darkRules(darkBranchVariable, "p2")[0]?.dv).toBe(1);
+  });
+});
+
+describe("an extra rule publishes its own inherited colour", () => {
+  const lightDarkColor = `.p3 { color: light-dark(red, blue); }`;
+
+  test("the light rule publishes the light colour", () => {
+    expect(variable(lightRule(lightDarkColor, "p3"), "__rn-css-color")).toBe(
+      "#f00",
+    );
+  });
+
+  /**
+   * Stated over every dark rule rather than over a count of them, because how
+   * many parses a colour declaration takes is not this rule's subject — that a
+   * dark rule never hands descendants the light colour is.
+   */
+  test("no dark rule publishes the light colour", () => {
+    const published = darkRules(lightDarkColor, "p3").map((rule) =>
+      variable(rule, "__rn-css-color"),
+    );
+
+    expect(published.length).toBeGreaterThan(0);
+    expect(published).not.toContain("#f00");
+  });
+
+  test("a dark rule publishes the dark colour", () => {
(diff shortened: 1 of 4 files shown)


YevheniiKotyrlo:

**UNFIXED —** In dark mode the two-declaration row paints RED text — the background's dark rule re-asserted the colour's LIGHT value.

**FIXED —** In dark mode the two-declaration row is BLUE on yellow. The currentcolor child still reads the stale light red — the stated limit.

| --- | --- |
|  |  |

Note on this pair: the two frames are near-identical **by design, and upstream of this package**. `@expo/metro-config` runs lightningcss with browserslist defaults that lower `light-dark()` to a `var()` pair before the compiler ever sees a `light-dark()` node, so the subject rows are empty on both builds. The control row (which carries no `light-dark()`) paints on both, which is what makes the empty subjects a reportable finding rather than a failed capture. This PR is still the right fix for the rule it changes — it simply is not observable through Expo's default Metro pipeline.

PR #425, opened 2026-08-15
fix(native): evaluate container and media conditions correctly on both axes and every operator

YevheniiKotyrlo:
Eight defects in the container-query and media-query condition evaluators. Every fix is mutation-proven — the defect is reintroduced, the red cases counted, then reverted — and each commit carries its own figures.

## The defects

1. **Every `min-width` / `max-width` container query was a strict greater-than.** `>=`, `<` and `<=` all returned `left > right` in `native/conditions/container-query.ts`; only `>` and `=` were correct. So `@container (max-width: 400px)` matched exactly when it should not.
2. **Every container reported its height as its width.** `containerHeightFamily` read `?.width`, so `@container (orientation: …)` answered `portrait` for every real container — `width > height` was unsatisfiable — and every height-based query answered on the wrong axis.
3. **`inline-size` and `block-size` never matched.** Both compile cleanly, and `getContainerFeatureValue` answered `undefined` for each, so `@container (min-inline-size: 400px)` matched nothing at any size. That is the axis `container-type: inline-size` names, which makes it the feature most container queries are written against. React Native lays out in one writing mode, so inline is the width axis and block the height axis.
4. **Interval (range-pair) conditions were never evaluated.** Both evaluators carried `case "[]": return false`, so `(400px < width < 800px)` was a no-op on both at-rules. The compiler emitted the pair faithfully throughout; only the evaluation was missing.
5. **`aspect-ratio` queries were dropped** instead of evaluated — `parseMediaFeatureValue` had no `ratio` arm, so the whole condition was unresolvable.
6. **A `<ratio>` with no finite quotient compiled to a value the bundle cannot carry.** `@media (min-aspect-ratio: 1/0)` produced `[">=", "aspect-ratio", Infinity]`. The native injection path serialises through `JSON.stringify`, which writes `Infinity` and `NaN` as `null`, so the condition meant one thing under jest and another on a device. A degenerate ratio is refused instead, which routes it into the same "did not compile" path as every other unresolvable value.
7. **A conditional block whose condition did not compile was kept** rather than dropped, so its declarations applied to every element carrying the class — the opposite of what the author wrote.
8. **Importing `react-native-css/compiler` booted the native runtime.** `compiler/compiler.types.ts` had a _value_ import of `VAR_SYMBOL`, and `verbatimModuleSyntax` means it is not elided, so the emitted `dist/commonjs/compiler/compiler.types.js` required native reactivity and registered `Dimensions` and `Appearance` listeners in a bundler process.

## One comparison, shared

`@media` and `@container` speak one `MediaFeatureComparison` vocabulary, and defect 1 is what a second hand-written copy of it costs. `native/conditions/compare.ts` now holds the only implementation, and both evaluators call it with whatever the feature answered rather than narrowing first — narrowing at the call site is exactly how the second copy gets written, since an evaluator that has to reject a keyword before it can call the primitive ends up deciding `=` itself.

Measured: making `=` wrong inside that primitive reddens **14** cases, six of them under `@container`, including the `orientation` keyword rows. Split across two implementations the same mutation reddened 8, and none of them under `@container`.

## Evidence

The figures below are the ones re-measured for the final pass, all under one rule: whole-suite, excluding the three pre-existing Windows `babel-plugin-tester` failures that are present on `main`.

| reintroduced defect | red |
| --- | --- |
| `=` decided wrongly in the shared comparison | 14 |
| the `<ratio>` arm removed | 17 |
| an interval's bounds swapped in the `@media` emit | 9 |
| `inline-size` / `block-size` back to unanswered | 5 |
| the finite-quotient check removed | 3 |
| the numeric guard removed from the shared comparison | 3 |
| an operator dropped from the shared census | 5 |

Two of those need a word.

- **The swapped-bounds emit.** 9 splits into 5 compiler-plane pins and 4 runtime rows. Against the commit that first evaluated intervals it is 4, because the compiler-plane `@media` interval table did not exist yet; against the base it is 0, because an unevaluated interval cannot observe its own layout. `@media` had no compiler-plane table at all for either the comparison operators or the interval layout, while `@container` was pinned twice — and one primitive now decides a range comparison for both, which is precisely where a divergence between their emits would land unseen.
- **An operator dropped from the census.** `src/__tests__/_media-features.ts` is one census (five operators, two axes, the `min-`/`max-` spellings) feeding five tables, and an incomplete census silently generates fewer cases while every suite stays green — deleting `"<"` removes 19 cases. Coverage is therefore asserted against `COMPARISON_MATCHES`, a total `Record` over `MediaFeatureComparison` whose keys are the union itself, not against the length of the generator each table was built from: that length is the census's length by construction and agrees with a census of any size, zero included.

Defect 8 is pinned from two sides. `src/__tests__/compiler/native-runtime-isolation.test.ts` walks the module graph and refuses any specifier out of `src/compiler/` that can reach `native/reactivity` on any platform — asking about reachability rather than about a list of runtime directory names, because the ways in are the package's own entry points: `react-native-css` re-exports `runtime`, which is `runtime.native` on native, and none of those three ids sits under a runtime directory. `src/__tests__/native/runtime-boot.test.ts` is the other side, loading a module into a fresh registry and counting what it attached to `Dimensions` and `Appearance`. Before writing it I measured that `babel-preset-expo` does _not_ elide an import whose only specifier sits in a type position — had it elided, the test would have passed for an unrelated reason.

Suite: **1072 → 1361 tests**, two runs with identical totals and zero suite-load failures. `yarn typecheck` and `yarn lint` exit 0. The 3 failures are the known Windows `babel-plugin-tester` baseline (an unrewritten relative `require("../View")`), present on `main`.

## Known limits, deliberately left standing

- **A comma-separated media query list is intersected, not unioned.** `rule.m` is a flat array fed from two places with opposite meanings — one entry per comma branch, which CSS unions, and one per enclosing `@media` block or media-carrying selector, which CSS intersects — and `testMediaQuery` intersects the whole array. Nesting is therefore right and a comma list is not: `@media (min-width: 400px), (min-height: 300px)` at 600x200 does not match, where CSS asks for either. Switching the evaluator to `.some(...)` fixes the list and breaks nesting, measured both ways; the emit has to say which kind of entry it is producing. That is a change to what is emitted rather than to how a condition is evaluated, so it is out of scope here and noted where the decision is made.
- **The boolean context is unimplemented.** `@media (width)` and `@container (width)` compile to `["!!", name]` and both evaluators answer `false`, so the query reads as valid and can never match. Answering it needs a truthiness rule per feature.
- `@container` has no counterpart to the `@media all` cases: `CompiledContainerCondition` `Exclude`s the `always` state, so it is unrepresentable rather than untested.

## Relationship to the media-condition-semantics branch

These two branches overlap and I would rather say so than have it surface at merge. They touch **10 files** in common, and `git merge-tree` against `fix/media-condition-semantics` reports **7 conflicted files with 17 conflict hunks**. No commit is duplicated — patch-id overlap is zero — but defects 1 and 2 are each fixed on both branches by _different_ implementations.

I am happy to rebase this onto whichever lands first, or to drop the overlapping commits from one side. Say which you prefer and I will restructure rather than ask you to resolve it.

### DIFF
--- a/src/__tests__/_media-features.ts
+++ b/src/__tests__/_media-features.ts
@@ -1,0 +1,155 @@
+import type { MediaFeatureComparison } from "react-native-css/compiler";
+
+/**
+ * The range vocabulary `@media` and `@container` share: the five comparison
+ * operators, the two size features, and what each operator means.
+ *
+ * It is shared rather than restated per suite because one meaning has to hold
+ * across the primitive, both evaluators and both at-rules. A second copy of a
+ * five-armed operator table is the exact shape of the defect these tests
+ * guard: two hand-written switches over the same five operators, differing by
+ * one character, one of them wrong.
+ *
+ * This is not a test file — `testPathIgnorePatterns` skips a path segment
+ * starting with an underscore.
+ */
+
+/**
+ * Where the measured value sits relative to the threshold the condition is
+ * written against. Every range comparison is decided by this and nothing else,
+ * so it is the dimension a table has to vary — and it is the dimension a
+ * copy-pasted operator arm hides in, because any two operators agree on at
+ * least one third of it.
+ */
+export type Ordering =
+  | "measured < threshold"
+  | "measured === threshold"
+  | "measured > threshold";
+
+export const ORDERINGS: Ordering[] = [
+  "measured < threshold",
+  "measured === threshold",
+  "measured > threshold",
+];
+
+/**
+ * What each comparison operator means, written out rather than computed.
+ *
+ * This is the specification every table is measured against. Deriving it from
+ * the code under test would make each table agree with whatever that code
+ * does, including a wrong operator — so it is literal, and it is the one place
+ * the semantics are stated.
+ *
+ * Typed as a total `Record`, so an operator added to `MediaFeatureComparison`
+ * is a compile error here rather than a silently uncovered arm.
+ */
+export const COMPARISON_MATCHES: Record<
+  MediaFeatureComparison,
+  Record<Ordering, boolean>
+> = {
+  "=": {
+    "measured < threshold": false,
+    "measured === threshold": true,
+    "measured > threshold": false,
+  },
+  ">": {
+    "measured < threshold": false,
+    "measured === threshold": false,
+    "measured > threshold": true,
+  },
+  ">=": {
+    "measured < threshold": false,
+    "measured === threshold": true,
+    "measured > threshold": true,
+  },
+  "<": {
+    "measured < threshold": true,
+    "measured === threshold": false,
+    "measured > threshold": false,
+  },
+  "<=": {
+    "measured < threshold": true,
+    "measured === threshold": true,
+    "measured > threshold": false,
+  },
+};
+
+export const COMPARISON_OPERATORS: MediaFeatureComparison[] = [
+  "=",
+  ">",
+  ">=",
+  "<",
+  "<=",
+];
+
+/**
+ * The `min-`/`max-` prefixed spelling of the two operators that have one.
+ *
+ * lightningcss normalises `(min-width: 400px)` into a `>=` range condition, so
+ * the prefixed form is not a separate feature — it is the same condition
+ * written a second way, and it has to compile to the same tuple and evaluate
+ * to the same verdict. It is also the spelling almost every author writes, so
+ * an operator defect reaches users through this row first.
+ */
+export const RANGE_PREFIX: Partial<
+  Record<MediaFeatureComparison, "min" | "max">
+> = {
+  ">=": "min",
+  "<=": "max",
+};
+
+/**
+ * The two size features a range condition is written against. Both at-rules
+ * accept both, and each has its own measurement — reading one axis off the
+ * other is a defect no single-axis table can see.
+ */
+export const SIZE_FEATURES = ["width", "height"] as const;
+
+export type SizeFeature = (typeof SIZE_FEATURES)[number];
+
+export interface SizeComparison {
+  /** The operator the runtime is handed, whatever spelling the CSS used. */
+  operator: MediaFeatureComparison;
+  feature: SizeFeature;
+  /** `range` is `(width >= 400px)`; `prefixed` is `(min-width: 400px)`. */
+  spelling: "range" | "prefixed";
+  /** The condition as written inside the query's parentheses. */
+  condition: (threshold: number) => string;
+  /** Test-name fragment, e.g. `width >=` or `min-width:`. */
+  label: string;
+}
+
+/**
+ * Every way to write a size range condition: each operator on each axis, plus
+ * the prefixed spelling of the two operators that have one.
+ */
+export function sizeComparisons(): SizeComparison[] {
+  return SIZE_FEATURES.flatMap((feature) => {
+    return COMPARISON_OPERATORS.flatMap((operator): SizeComparison[] => {
+      const prefix = RANGE_PREFIX[operator];
+
+      const range: SizeComparison = {
+        operator,
+        feature,
+        spelling: "range",
+        condition: (threshold) => `(${feature} ${operator} ${threshold}px)`,
+        label: `${feature} ${operator}`,
+      };
+
+      if (!prefix) {
+        return [range];
+      }
+
+      return [
+        range,
+        {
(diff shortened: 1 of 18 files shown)


YevheniiKotyrlo:
Cross-referencing #425 and #426, which conflict — flagging it now rather than at merge time.

They are independent siblings (`git merge-base` is `main`; neither is an ancestor of the other) and `git merge-tree` reports content conflicts in four files: `src/compiler/compiler.ts`, `src/compiler/container-query.ts`, `src/__tests__/compiler/media-query.test.ts`, `src/__tests__/native/container-queries.test.tsx`.

The hazard is in the `[]` (interval) arm of `testContainerQuery`, because a plausible resolution silently reverts work:

- **#425** evaluates the interval — `testMediaFeatureInterval(condition, value)` — and also maps `inline-size` / `block-size`.
- **#426** returns `UNKNOWN` unconditionally and never evaluates it, because at the time it was written an unorderable bound had no safe answer.

Taking #426's arm wholesale therefore un-fixes #425's range container queries. Taking #425's wholesale loses #426's three-valued handling, where `false` is the one answer a negation turns into a match.

The resolution that is correct in both directions evaluates the interval **and** returns `UNKNOWN` only when a bound is unorderable.

One more line worth keeping whichever way it lands: #426's boolean-context arm uses `isTruthyFeatureValue` (`Number.isFinite(value) && value !== 0`), and the plain `value !== 0` form admits `NaN`, `Infinity` and `-Infinity`. That is reachable — `@container (aspect-ratio)` compiles to `["!!","aspect-ratio"]` and the feature value is `width / height`, so a zero-height container yields `Infinity` and the weaker form matches where it should not.

A two-case test pins both halves, and neither branch passes it alone:

1. `@container box (400px < width < 800px)` over a 500x100 container must match. (#425 ✅ · #426 ❌)
2. `@container box (10em < width < 20em)` and its `not (…)` twin must **both** fail over the same container — an unorderable bound is unknown, not false. (#425 ❌, its negation matches · #426 ✅)

YevheniiKotyrlo:
Correction to my note above — I undercounted the conflict set, and the omission was the load-bearing file.

`git merge-tree --write-tree fix/container-query-defects fix/media-condition-semantics` reports **seven** conflicted files, not four:

```
src/compiler/compiler.ts
src/compiler/container-query.ts
src/compiler/media-query.ts                  ← omitted above
src/native/conditions/container-query.ts     ← omitted above
src/native/conditions/media-query.ts         ← omitted above
src/__tests__/compiler/media-query.test.ts
src/__tests__/native/container-queries.test.tsx
```

`src/native/conditions/container-query.ts` is the file whose `[]`-arm resolution my note describes, so leaving it off the list made the note harder to act on rather than merely incomplete.

**A second resolution needs the same care, in `src/compiler/container-query.ts`'s `"not"` arm.** `main` has `return query ? ["!", query] : undefined`. #425 changes `parseContainerCondition` to return a discriminated union, so its arm becomes `query.type === "condition" ? ["!", query.condition] : undefined` — which DROPS an unrepresentable operand. #426 keeps main's return type and writes `return ["!", query ?? ["?"]]`, so an unrepresentable operand negates to unknown (MQ5 §3.1).

Taking #426's arm onto #425's types does not compile — `query` is an object there and never nullish. Taking #425's arm silently reverts #426's fix. The resolution that keeps both is #425's discriminant with #426's fallback:

```ts
["!", query.type === "condition" ? query.condition : (["?"] as MediaCondition)]
```

**And one more, in `src/native/conditions/compare.ts`.** #426 introduces `MediaFeatureOperand = Exclude<StyleDescriptor, undefined> | null` and a `parseMediaFeatureOperand` that can hand the comparator `null`; `compare.ts` exists only on #425, where both operands are typed `StyleDescriptor`, which excludes `null`. On the merge those call sites are a compile error rather than a silent `null === null → true` — so it self-flags, but the fix is to widen both operands and guard `left !== null` before the equality shortcut.

YevheniiKotyrlo:

**UNFIXED —** Only the control paints. Three of the five range operators collapsed onto `>`, and the height axis read the container WIDTH, so both subjects answer a comparison nobody wrote.

**FIXED —** All three bars paint. Every operator means what CSS says it means, and a height query reads the height.

| --- | --- |
|  |  |

PR #426, opened 2026-08-15
fix(native): answer an undecidable media condition with unknown rather than false

YevheniiKotyrlo:
The media- and container-condition evaluators answer several questions wrongly, and one of them wrongly in a way that cannot be distinguished from a legitimate "no".

## The core problem: a condition the runtime cannot decide answered `false`

An operand that failed to resolve was treated as a negative rather than as unknown. `false` and "I could not work this out" are different facts, and collapsing them means a rule silently does not apply and nothing anywhere reports why. This branch introduces the unknown state, carries the unresolved-operand marker in a shape JSON can actually hold, and refuses an operand that could not be resolved rather than guessing.

## The rest

- **A comma-separated media query list did not compile as a union.** `@media (a), (b)` is a disjunction; one refused branch dropped the whole list.
- **Boolean media and container features were not evaluated at all** — the bare `(hover)` / `(pointer)` form, as distinct from `(hover: hover)`.
- **`hover`, `color` and `prefers-color-scheme` were answered from different places.** They now come from one source, which is what stops them disagreeing.
- **Every container range operator now has its own comparison** — `>=`, `<` and `<=` previously all evaluated as `left > right`.
- **A container reported its height as its width**, so orientation and every height query answered on the wrong axis.

## Evidence

A reviewer pair audited this branch and converged on SHIP-WITH-CHANGES; the one point they disagreed on was settled by measurement rather than by argument, and the changes they asked for are in.

The last two commits are worth reading as a pair: one pins the two condition arms no test observed, and the next deliberately hands the media-prelude backstop to the branch that actually tests it, rather than keeping a second copy here.

## Relationship to the container-query-defects branch

I am submitting both, and they overlap — better said here than discovered at merge. They share **10 files** and `git merge-tree` reports **9 conflict sites**.

No commit is duplicated between them: patch-id overlap is **zero**. But two defects — the container range operators, and the height axis — are fixed on *both* branches by *different* implementations. That is my doing, from working the two problems in parallel.

Tell me which you would rather take first and I will rebase the other onto it, or collapse the overlapping commits into one branch. I would rather restructure than hand you nine conflicts.

### DIFF
--- a/src/__tests__/compiler/media-query.test.ts
+++ b/src/__tests__/compiler/media-query.test.ts
@@ -1,4 +1,15 @@
+import type { MediaCondition } from "react-native-css/compiler";
 import { compile } from "react-native-css/compiler";
+
+/** The media conditions of every rule compiled for `className`. */
+function mediaConditions(css: string, className: string) {
+  const rules =
+    compile(css)
+      .stylesheet()
+      .s?.find(([name]) => name === className)?.[1] ?? [];
+
+  return rules.map((rule): MediaCondition[] | undefined => rule.m);
+}
 
 describe.skip("platform media queries", () => {
   test("android", () => {
@@ -59,6 +70,76 @@ describe.skip("platform media queries", () => {
         ],
       ],
     });
+  });
+});
+
+describe("comma-separated media query lists", () => {
+  test("compile to a union, not an intersection", () => {
+    expect(
+      mediaConditions(
+        `@media (min-width: 100px), (min-width: 9999px) {
+          .my-class { background-color: red; }
+        }`,
+        "my-class",
+      ),
+    ).toStrictEqual([
+      [
+        [
+          "|",
+          [
+            [">=", "width", 100],
+            [">=", "width", 9999],
+          ],
+        ],
+      ],
+    ]);
+  });
+
+  test("a single query is not wrapped", () => {
+    expect(
+      mediaConditions(
+        `@media (min-width: 100px) {
+          .my-class { background-color: red; }
+        }`,
+        "my-class",
+      ),
+    ).toStrictEqual([[[">=", "width", 100]]]);
+  });
+
+  test("a comma list and an `or` condition compile identically", () => {
+    const comma = mediaConditions(
+      `@media (min-width: 100px), (min-width: 9999px) {
+        .my-class { background-color: red; }
+      }`,
+      "my-class",
+    );
+
+    const or = mediaConditions(
+      `@media ((min-width: 100px) or (min-width: 9999px)) {
+        .my-class { background-color: red; }
+      }`,
+      "my-class",
+    );
+
+    expect(comma).toStrictEqual(or);
+  });
+
+  test("nested @media rules still intersect", () => {
+    expect(
+      mediaConditions(
+        `@media (min-width: 100px) {
+          @media (min-height: 200px) {
+            .my-class { background-color: red; }
+          }
+        }`,
+        "my-class",
+      ),
+    ).toStrictEqual([
+      [
+        [">=", "width", 100],
+        [">=", "height", 200],
+      ],
+    ]);
   });
 });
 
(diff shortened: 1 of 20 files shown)


YevheniiKotyrlo:
Cross-referencing #425 and #426, which conflict — flagging it now rather than at merge time.

They are independent siblings (`git merge-base` is `main`; neither is an ancestor of the other) and `git merge-tree` reports content conflicts in four files: `src/compiler/compiler.ts`, `src/compiler/container-query.ts`, `src/__tests__/compiler/media-query.test.ts`, `src/__tests__/native/container-queries.test.tsx`.

The hazard is in the `[]` (interval) arm of `testContainerQuery`, because a plausible resolution silently reverts work:

- **#425** evaluates the interval — `testMediaFeatureInterval(condition, value)` — and also maps `inline-size` / `block-size`.
- **#426** returns `UNKNOWN` unconditionally and never evaluates it, because at the time it was written an unorderable bound had no safe answer.

Taking #426's arm wholesale therefore un-fixes #425's range container queries. Taking #425's wholesale loses #426's three-valued handling, where `false` is the one answer a negation turns into a match.

The resolution that is correct in both directions evaluates the interval **and** returns `UNKNOWN` only when a bound is unorderable.

One more line worth keeping whichever way it lands: #426's boolean-context arm uses `isTruthyFeatureValue` (`Number.isFinite(value) && value !== 0`), and the plain `value !== 0` form admits `NaN`, `Infinity` and `-Infinity`. That is reachable — `@container (aspect-ratio)` compiles to `["!!","aspect-ratio"]` and the feature value is `width / height`, so a zero-height container yields `Infinity` and the weaker form matches where it should not.

A two-case test pins both halves, and neither branch passes it alone:

1. `@container box (400px < width < 800px)` over a 500x100 container must match. (#425 ✅ · #426 ❌)
2. `@container box (10em < width < 20em)` and its `not (…)` twin must **both** fail over the same container — an unorderable bound is unknown, not false. (#425 ❌, its negation matches · #426 ✅)

YevheniiKotyrlo:
Correction to my note above — I undercounted the conflict set, and the omission was the load-bearing file.

`git merge-tree --write-tree fix/container-query-defects fix/media-condition-semantics` reports **seven** conflicted files, not four:

```
src/compiler/compiler.ts
src/compiler/container-query.ts
src/compiler/media-query.ts                  ← omitted above
src/native/conditions/container-query.ts     ← omitted above
src/native/conditions/media-query.ts         ← omitted above
src/__tests__/compiler/media-query.test.ts
src/__tests__/native/container-queries.test.tsx
```

`src/native/conditions/container-query.ts` is the file whose `[]`-arm resolution my note describes, so leaving it off the list made the note harder to act on rather than merely incomplete.

**A second resolution needs the same care, in `src/compiler/container-query.ts`'s `"not"` arm.** `main` has `return query ? ["!", query] : undefined`. #425 changes `parseContainerCondition` to return a discriminated union, so its arm becomes `query.type === "condition" ? ["!", query.condition] : undefined` — which DROPS an unrepresentable operand. #426 keeps main's return type and writes `return ["!", query ?? ["?"]]`, so an unrepresentable operand negates to unknown (MQ5 §3.1).

Taking #426's arm onto #425's types does not compile — `query` is an object there and never nullish. Taking #425's arm silently reverts #426's fix. The resolution that keeps both is #425's discriminant with #426's fallback:

```ts
["!", query.type === "condition" ? query.condition : (["?"] as MediaCondition)]
```

**And one more, in `src/native/conditions/compare.ts`.** #426 introduces `MediaFeatureOperand = Exclude<StyleDescriptor, undefined> | null` and a `parseMediaFeatureOperand` that can hand the comparator `null`; `compare.ts` exists only on #425, where both operands are typed `StyleDescriptor`, which excludes `null`. On the merge those call sites are a compile error rather than a silent `null === null → true` — so it self-flags, but the fix is to widen both operands and guard `left !== null` before the equality shortcut.

YevheniiKotyrlo:

**UNFIXED —** BOTH bars paint. The unimplemented feature answers `false`, and `not false` is true — so a rule guarded by a condition the runtime cannot decide applies anyway.

**FIXED —** Only the control paints. An undecidable condition answers `unknown`, and `not unknown` is still unknown, so the negated rule correctly does not apply.

| --- | --- |
|  |  |

PR #427, opened 2026-08-15
fix(native): stop a read swallowing a notification, and share reactivity state across a dual-package split

YevheniiKotyrlo:
Five fixes to the native reactivity primitive and the registries built on it.

## The defects

1. **A read could swallow an observable notification.** The cache advanced past the value a subsequent `run()`'s guard would read, so a subscriber that read at the wrong moment never learned the value had changed. This is the primitive every `dark:`, `hover:` and container-query rule is answered from, so a swallowed notification is a rule that silently stops re-evaluating.
2. **`reactivity.ts` had no dual-package guard at all.** Its process-global state is now pinned to `globalThis`, so two copies of the module — the CJS and ESM builds loaded by different consumers — share one registry rather than each keeping a private one that the other's writes never reach.
3. **`universalVariables` was created and exported but never written.** It now injects into its own family. Measured before fixing: universal correctly outranks root, so this was dead code plus an accidental ordering rather than a wrong value — the fix makes the intent real rather than repairing a live wrong answer.
4. **The jest harness leaked root and universal variables between tests.** `beforeEach` cleared `StyleCollection.styles` but not `rootVariables` or `universalVariables`, so a `vr` entry survived into the next test. That is a harness defect that makes other people's tests pass for the wrong reason, and it already forced one `:root` test into a file of its own.
5. A `:root` variable test now reaches the runtime registry rather than asserting only the compiled shape.

## Evidence

Each fix has a red-then-green test. Fix 1's is the interesting one: it drives the primitive directly rather than through a rule, because a notification swallowed inside a rule evaluation is indistinguishable from a rule that legitimately did not match.

Fix 4 is worth calling out as a harness change rather than a product change — it makes some existing tests stricter, and any that were relying on the leak are now honest.

`yarn typecheck` and `yarn lint` exit 0. The 3 remaining failures are the known Windows `babel-plugin-tester` baseline (an unrewritten relative `require("../View")`), present on `main` and unrelated.

### DIFF
--- a/src/__tests__/native/reactivity-dual-package.test.ts
+++ b/src/__tests__/native/reactivity-dual-package.test.ts
@@ -1,0 +1,132 @@
+import type { Effect } from "../../native/reactivity";
+
+/**
+ * `package.json`'s `exports` map sends `import` to `dist/module/**` and
+ * `require` to `dist/commonjs/**`, and Metro resolves per requesting module, so
+ * a single app can hold two copies of this module. Two copies means two sets of
+ * observables and two batches, and a write through one is invisible to the
+ * other.
+ *
+ * `jest.resetModules()` followed by a fresh `import()` reproduces that exactly:
+ * the source is evaluated twice against one `globalThis`. It is also fully
+ * deterministic - module evaluation is synchronous and ordered, so there is no
+ * clock, no listener race and nothing carried between tests.
+ */
+type ReactivityModule = typeof import("../../native/reactivity");
+
+async function loadCopy(): Promise<ReactivityModule> {
+  jest.resetModules();
+
+  return import("../../native/reactivity");
+}
+
+function createSubscriber(): Effect & { runs: number } {
+  const subscriber = {
+    observers: new Set<Effect>(),
+    runs: 0,
+    run: () => {
+      subscriber.runs++;
+    },
+  };
+
+  return subscriber;
+}
+
+describe("dual package hazard", () => {
+  test("a second evaluation really is a separate copy", async () => {
+    const first = await loadCopy();
+    const second = await loadCopy();
+
+    // Vacuity guard. If the module were not re-evaluated, every sharing
+    // assertion below would hold for the wrong reason. `observable` is a plain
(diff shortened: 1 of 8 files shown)


YevheniiKotyrlo:

**UNFIXED —** The first two bars are red. `containerHeightFamily` projects `.width` off the shared layout cell, so `@container (height >= 60px)` is answered with the container's width — 96 clears 60 and the rule applies.

**FIXED —** Only the second bar is red. The height query reads the host's height — 24, which does not clear 60 — so the rule is withheld and the bar matches the no-container baseline.

| before — this build, minus this PR | after — with this PR |
| --- | --- |
|  |  |

Read the `container-axis` row. One host governs the first two bars — `container-type: size`, 96×24, deliberately NOT square — and one threshold governs both queries: `@container (height >= 60px)` on the first, `@container (width >= 60px)` on the second. 96 clears 60 and 24 does not, so a single field cannot satisfy both.

**The second and third bars are the anti-vacuity pair, and they do not move.** The second is red in both frames, so the container registered and size queries resolve at all; the third sits under no container and is blue in both, so a subject with nothing above it was never matching. Only the first — the height query — flips.

That is also why a square host would have hidden this indefinitely: with the two axes equal, the wrong field returns a plausible number and every assertion downstream passes on it. Every other container case in the gallery used one, which is why nothing caught it.

Measured rather than asserted: **72 of the two frames' 2400 pixel rows differ, and they are rows 1308–1379** — exactly the bar band `uiautomator` reports for these three elements. The label above them and all six other rows are byte-identical, because `before` is this same build with only this PR's hunk reverted in both `dist/commonjs` and `dist/module`, so exactly one variable differs.

Two more consumers move with it, neither visible here: `getContainerFeatureValue`'s `orientation` arm computes `width > height`, which became `width > width` and reported `portrait` for every container, and its `aspect-ratio` arm computed `width / width` and reported exactly `1` whatever the box was.

Both frames: pooled Android 16 emulator, 1140×2400 @ 480dpi, light scheme, same session.

PR #428, opened 2026-08-15
fix: six small findings, plus a windows CI lane

YevheniiKotyrlo:
Six small independent findings, each separable — split any of them out if you would rather take them apart.

## Fixes

1. **`prefers-color-scheme` resolved to nothing when the OS reported no preference.** It now resolves to `light`, which is what the CSS spec says `no-preference` means for this feature. Without it a `dark:` rule and its unprefixed counterpart could both fail to apply.
2. **`src/web/api.tsx` called a method that does not exist.** `react-native-web@0.21.1`'s `Appearance` exports only `getColorScheme` and `addChangeListener` — there is no `setColorScheme`, so `colorScheme.set(…)` was a hard `TypeError` on web. It is invisible to `tsc` because the file imports `Appearance` from `"react-native"`, so TypeScript resolves React Native's own `.d.ts` while the bundler substitutes `react-native-web`. It now reports that the scheme cannot be set on web rather than throwing.

## Tests, and one CI lane

3. **`ci: run the unit suite on windows.`** `ci.yml` was `ubuntu-latest` plus one `macos-15`, with no Windows runner — which is why a three-test failure sat in `main` unnoticed and why a previous cross-platform test of mine was inert on the only runner that existed. This is the smallest change here and probably the most valuable.
4. **`prefers-color-scheme` had no compiler test.** `media-query.test.ts` had no case for it — its platform block is `describe.skip`, leaving only `hover`. The single existing assertion of that condition reached it via `light-dark()`, where the compiler *synthesises* the condition, so the `@media` parse path was unproven. The platform media queries are revived in the same commit.
5. **`src/metro/` had no tests at all.** `metro/resolver.ts` is the only import rewriter when `globalClassNamePolyfill` is false, and the babel plugin the only one when it is true — they are alternatives, not layers, so one of the two rewriting paths was entirely unexercised. This covers the gate.
6. **`dynamicRootVariables`** so a `:root` test reaches the runtime rather than asserting only the compiled output.

## Note

Item 2 is a behaviour change on web for anyone currently calling `colorScheme.set` there — today that call throws, so nothing can be depending on it working, but it is the one item here that is not purely additive. Happy to split it out if you would rather take it separately.

`yarn typecheck` and `yarn lint` exit 0. The 3 remaining failures are the known Windows `babel-plugin-tester` baseline, present on `main`.

### DIFF
--- a/.claude/skills/add-test/SKILL.md
+++ b/.claude/skills/add-test/SKILL.md
@@ -41,6 +41,12 @@ Verify CSS → JSON compilation output structure.
 
 Test runtime style application on native platform.
 
+> A `:root` custom property with exactly one declaration is inlined by the
+> `inlineVariables` compiler pass and never reaches the runtime, so a test
+> written that way passes with the runtime registry deleted. Declare it with
+> `dynamicRootVariables` from `react-native-css/jest`. See the Testing section
+> of `DEVELOPMENT.md`.
+
 ## Steps
 
 1. **Identify the feature**: What needs testing? Use `$ARGUMENTS` as the starting point.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,22 @@ jobs:
       - name: Run unit tests
         run: yarn test --maxWorkers=2 --coverage
 
+  # The babel plugin, the metro resolver and the compiler all join and compare
+  # file paths. A separator or drive-letter bug in any of them is invisible to a
+  # POSIX runner and breaks every Windows contributor, so the unit suite runs
+  # here too. Coverage is left to the ubuntu job; this one only needs to fail.
+  test-windows:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Setup
+        uses: ./.github/actions/setup
+
+      - name: Run unit tests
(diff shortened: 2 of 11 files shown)


YevheniiKotyrlo:

Five of the six findings are test and CI coverage, and the sixth — `prefers-color-scheme` resolving to `light` when the host states no preference — changes which rule applies rather than how a frame is painted, so a pair would be two identical screens on any device whose OS reports a preference.

What is worth measuring instead is the finding that made the rest reachable. `src/web/api.tsx` called `Appearance.setColorScheme`, which `react-native-web@0.21.1` does not export, so `colorScheme.set(…)` was a hard `TypeError` on web. It is invisible to `tsc` because the file imports `Appearance` from `"react-native"` — TypeScript resolves React Native's own `.d.ts` while the bundler substitutes `react-native-web`, so the type and the runtime disagree and only the runtime is right.

Gate results at this branch's head, on Windows:

```
tsc --noEmit   0 errors
eslint         0 problems
jest           1074 passed, 3 failed, 19 skipped
```

The three are the Windows-only `babel-plugin-tester` baseline. **The CI lane this PR adds is what would surface them upstream** rather than only on a contributor's machine — `ci.yml` was `ubuntu-latest` plus one `macos-15`, with no Windows runner, which is why a three-test failure sat in `main` unnoticed.

PR #429, opened 2026-08-15
fix: compile on every react-native in the peer range, not just 0.81

YevheniiKotyrlo:
`peerDependencies` says `react-native: ">=0.81"`. **0.81 is the only minor in that range the package compiles on.** Measured with `yarn typecheck` at `f70c402`, one install per version:

| `react-native` | errors |
| --- | --- |
| 0.81.4 | **0** |
| 0.82.0 | **6** |
| 0.83.0 | **7** |
| 0.84.0 | **7** |
| 0.85.3 | **7** |
| 0.86.0 | **7** |

Two unrelated causes, with different start points, plus a jest bootstrap failure that makes the suite unrunnable on 0.86. This PR makes all six compile and keeps 0.81 exactly where it is.

The seven on 0.86:

```
src/components/FlatList.tsx(12,16): error TS2590: Expression produces a union type that is too complex to represent.
src/jest/index.ts(24,29): error TS2345: Argument of type 'null' is not assignable to parameter of type 'ColorSchemeName'.
src/jest/index.ts(25,19): error TS2345: Argument of type 'null' is not assignable to parameter of type 'Value'.
src/native/api.tsx(76,31): error TS2345: Argument of type 'string' is not assignable to parameter of type 'Value'.
src/native/reactivity.ts(220,3): error TS2345: Argument of type 'ColorSchemeName | null | undefined' is not assignable to parameter of type 'ColorSchemeName | Read<ColorSchemeName, ColorSchemeName>'.
src/native/reactivity.ts(222,57): error TS2345: Argument of type 'string' is not assignable to parameter of type 'Value'.
src/web/api.tsx(72,3): error TS2322: Type '() => ColorSchemeName | null | undefined' is not assignable to type '() => ColorSchemeName'.
```

## 1. `ColorSchemeName` inverts at 0.82 (6 errors, 0.82 and later)

`react-native`'s own type changes meaning inside the declared range:

| | `ColorSchemeName` | `Appearance.setColorScheme` accepts | `null` at runtime |
| --- | --- | --- | --- |
| 0.81 | `"light" \| "dark" \| null \| undefined` | `ColorSchemeName \| null \| undefined` | mapped to `"unspecified"` |
| 0.82+ | `"light" \| "dark" \| "unspecified"` | `ColorSchemeName` — `null` rejected | passed straight through |

`runtime.types.ts` re-exported that type as `ColorScheme`'s parameter and return, so the package's public API changed shape with the installed `react-native`, and the implementation could not satisfy both ends. The four errors reporting a bare generic `Value` are cascades of the failed `observable<ColorSchemeName>(...)` call at `reactivity.ts:220`.

The two `setColorScheme` parameter types overlap only on `"light" | "dark"`, so **no literal naming "follow the system" type-checks on both**. The runtime does not split the same way: `"unspecified"` is what the native module receives on every version, because 0.81 maps `null` to it on the way in.

`ColorSchemeName` is now the package's own union of every spelling, `"light" | "dark" | "unspecified" | null | undefined`. Both directions widen, so nothing a caller passes today stops working, and `"unspecified"` becomes writable on 0.81 too. The one incompatible call goes through `setAppearanceColorScheme`: `"light"` and `"dark"` keep the shared declaration, everything else is written as `"unspecified"` through a locally restated signature — one place, reason inline, covering only the value the versions disagree about.

## 2. `DotNotation` walks into component instances (TS2590, 0.83 and later)

Different cause, so worth reading separately.

`ScrollViewProps.scrollViewRef` is a `RefObject<ScrollView>`, and `ScrollView` carries `ScrollViewProps` again — the prop graph `DotNotation` enumerates is cyclic. Bounded at `MaxDepth 10`, that one prop contributes **4,654 paths on 0.81**. From 0.83 the host instance surface is wide enough that building the union crosses TypeScript's 100,000-member limit, so `StyledConfiguration<typeof FlatList>` fails and takes `src/components/FlatList.tsx` with it. `FlatList.d.ts` is unchanged across all of these versions; the growth is underneath it.

None of those paths can ever be a mapping target — they run through an object that does not exist until render. Treating a class component instance as a leaf removes the cycle at its source. Lowering `MaxDepth` would only move the threshold to whichever prop type grows next.

Path-set diff on 0.81, where the old and new definitions both resolve:

| | before | after | removed | added |
| `ViewProps` | 733 | 733 | 0 | 0 |
| `ScrollViewProps` | 7,482 | 2,094 | 5,388 | 0 |
| `FlatListProps<unknown>` | 9,204 | 3,816 | 5,388 | 0 |
| `Pick<ScrollViewProps, "scrollViewRef">` | 4,656 | 2 | 4,654 | 0 |

Every removed path traverses a ref (`.current.`). Nothing is added.

## 3. The suite cannot bootstrap on 0.86

Independent of the types, and a prerequisite for testing any of this. `jest-expo@56` declares `@react-native/jest-preset` as a **non-optional** peer and resolves `setupFiles` and `testEnvironment` through it. Nothing provided it, so the suite ran on the copy `react-native` shipped in-tree. 0.86 deleted that copy; its `jest-preset.js` is now a shim that throws, and `.config/jest.config.cjs` dies before a single test loads:

```
The React Native Jest preset that jest-expo relies on has moved to a separate package.
    at Object.<anonymous> (node_modules/jest-expo/jest-preset.js:16:13)
    at Object.<anonymous> (.config/jest.config.cjs:4:18)
```

The package has no 0.81.x release — 0.85.0 is the first — so it cannot be pinned to this repo's `react-native`. `^0.85.0` matches jest-expo's own declared range and works at both ends. On 0.81.4 it is a no-op: the suite is bit-identical with and without it.

## Verification

**Typecheck, every minor in the range**, one install each, root and `example/` manifests moved together:

| | 0.81.4 | 0.82.0 | 0.83.0 | 0.84.0 | 0.85.3 | 0.86.0 |
| --- | --- | --- | --- | --- | --- | --- |
| `upstream/main` | 0 | 6 | 7 | 7 | 7 | 7 |
| this branch | **0** | **0** | **0** | **0** | **0** | **0** |

**Lint, build and the full suite at both ends of the range:**

| | `react-native@0.81.4` | `react-native@0.86.0` |
| --- | --- | --- |
| `yarn lint` | clean | clean |
| `yarn build` | ok | ok |
| `yarn test` | `3 failed, 21 skipped, 1055 passed, 1079 total` | `3 failed, 21 skipped, 1055 passed, 1079 total` |

Identical, with `numRuntimeErrorTestSuites: 0` on each — every suite loaded. The 3 failures are pre-existing on `main` before this branch: `babel-plugin-tester` path mismatches over an unrewritten relative `require("../View")`, reproducing on Windows only. Every figure above is the second of two consecutive runs, since a cold first run in a fresh tree can drop suites and silently subtract their whole count.

Running the suite on 0.86 also needs `@react-native/babel-preset` at 0.86 — `babel-preset-expo@54` pins 0.81.4, whose codegen plugin cannot parse 0.86's native component specs (`Unable to determine event arguments for "onChange"`, 21 suites down). That is an Expo SDK pairing issue rather than anything in this package, so it is not in the diff; I pinned it through `resolutions` locally for the measurement only.

## Tests

Both fixes are type-surface, so each carries a compile-plane assertion, and the color scheme one carries a runtime assertion as well.

`src/__tests__/native/color-scheme.test.tsx`

- Compile plane: three aliases assert `ColorSchemeName` covers `react-native`'s own union, `Appearance.getColorScheme()`'s return, and the change-listener payload. They are read off the **installed** `react-native` rather than restating either spelling, so `yarn typecheck` fails at whichever version the checkout is on the moment coverage lapses.
- Runtime plane: `NativeAppearance` is mocked so writes are observable — it is `TurboModuleRegistry.get('Appearance')`, null under jest on every version, which is why `setColorScheme` was previously unobservable in tests. Five cases assert what the native module actually receives; a sixth asserts `@media (prefers-color-scheme: dark)` engages on `"dark"` and releases on `"unspecified"`.

`src/__tests__/native/components.test.tsx`

- Compile plane: `scrollViewRef` and `scrollViewRef.current` are reachable targets, and a `@ts-expect-error` pins that `scrollViewRef.current.props.style` is not. That directive is what makes the guard work on 0.81 and 0.82 too, where TS2590 never fires — reverting the fix there fails with `TS2578: Unused '@ts-expect-error' directive`.
- No runtime plane: `dot-notation.types.ts` compiles to `export {}` (72 bytes), so it has no runtime surface. `FlatList`'s mapping behaviour is already covered by `className-with-style.test.tsx`, which passes at both ends.

Each fix was mutation-proved by reverting it and watching the guard go red:

| Reverted | 0.81.4 | 0.86.0 |
| --- | --- | --- |
| component-instance leaf | `TS2578` unused `@ts-expect-error` | `TS2578` + the original `FlatList.tsx(12,16) TS2590` |
| package-owned `ColorSchemeName` | `TS2345` on `colorScheme.set("unspecified")` | that plus the original `web/api.tsx TS2322` |
| `setAppearanceColorScheme` sends the raw value | **passes** — 0.81 maps `null` itself, so this version cannot see it | 2 tests red: native module receives `null` / `undefined` |
| `@react-native/jest-preset` devDependency | no effect — in-tree preset still resolves | jest config fails to load, whole suite unrunnable |

## Out of scope, noticed while here

`react-native-web@0.21`'s `Appearance` exports `getColorScheme` and `addChangeListener` but no `setColorScheme`, so `colorScheme.set()` on web throws `TypeError` at runtime today. Unrelated to this, and unchanged by this PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,6 +215,7 @@
     "@eslint/js": "^10.0.1",
     "@expo/metro-config": "~54.0.5",
     "@ianvs/prettier-plugin-sort-imports": "^4.4.2",
+    "@react-native/jest-preset": "^0.85.0",
     "@release-it/conventional-changelog": "10.0.1",
     "@tailwindcss/postcss": "^4.1.12",
     "@testing-library/react-native": "^13.3.3",
--- a/src/__tests__/native/color-scheme.test.tsx
+++ b/src/__tests__/native/color-scheme.test.tsx
@@ -1,0 +1,105 @@
+import {
+  Appearance,
+  type ColorSchemeName as ReactNativeColorSchemeName,
+} from "react-native";
+
+import { act, render, screen } from "@testing-library/react-native";
+import { View } from "react-native-css/components/View";
+import { registerCSS, testID } from "react-native-css/jest";
+import { colorScheme } from "react-native-css/runtime";
+import type { ColorSchemeName } from "react-native-css/runtime.types";
+
+import { setAppearanceColorScheme } from "../../color-scheme";
+
+const mockSetColorScheme = jest.fn();
+
+// `Appearance` reads its native module lazily, so mocking the module is enough to make
+// the writes observable. Without it `TurboModuleRegistry.get('Appearance')` is null
+// under jest and every `setColorScheme` is a silent no-op.
+jest.mock("react-native/Libraries/Utilities/NativeAppearance", () => ({
+  __esModule: true,
+  default: {
+    setColorScheme: (scheme: string) => {
+      mockSetColorScheme(scheme);
+    },
+    // `getColorScheme` feeds react-native's own post-write bookkeeping, which asserts
+    // the value is a resolved scheme or absent.
+    getColorScheme: () => undefined,
+    addListener: () => undefined,
+    removeListeners: () => undefined,
+  },
+}));
+
+/*******************************  Compile plane  ******************************/
+
+type Assert<T extends true> = T;
+type Covers<Actual, Accepted> = [Actual] extends [Accepted] ? true : false;
+
+type AppearanceChangeColorScheme = Parameters<
+  Parameters<typeof Appearance.addChangeListener>[0]
+>[0]["colorScheme"];
+
+/**
+ * `ColorSchemeName` is react-native-css's own union rather than react-native's, because
+ * react-native's changes shape inside this package's peer range: up to 0.85 "follow the
+ * system" is `null | undefined`, from 0.86 it is `"unspecified"`.
+ *
+ * These aliases are read off the installed react-native rather than restating either
+ * spelling, so `yarn typecheck` fails at whichever end of the range the checkout is on
+ * the moment the package stops covering a value react-native hands it.
+ */
+export type CoversReactNativeUnion = Assert<
+  Covers<ReactNativeColorSchemeName, ColorSchemeName>
+>;
+export type CoversAppearanceRead = Assert<
+  Covers<ReturnType<typeof Appearance.getColorScheme>, ColorSchemeName>
+>;
+export type CoversAppearanceEvent = Assert<
+  Covers<AppearanceChangeColorScheme, ColorSchemeName>
+>;
+
+/*******************************  Runtime plane  ******************************/
+
+beforeEach(() => {
+  mockSetColorScheme.mockClear();
+});
+
+test.each([
+  ["light", "light"],
+  ["dark", "dark"],
+  ["unspecified", "unspecified"],
+  [null, "unspecified"],
+  [undefined, "unspecified"],
+] satisfies [ColorSchemeName, string][])(
+  "setAppearanceColorScheme(%p) reaches the native module as %p",
+  (value, expected) => {
+    setAppearanceColorScheme(value);
+
+    expect(mockSetColorScheme).toHaveBeenCalledTimes(1);
+    expect(mockSetColorScheme).toHaveBeenCalledWith(expected);
+  },
+);
+
+test("prefers-color-scheme follows the scheme and releases on 'unspecified'", () => {
+  registerCSS(`
+.my-class { color: blue; }
+
+@media (prefers-color-scheme: dark) {
+  .my-class { color: red; }
+}`);
+
+  render(<View testID={testID} className="my-class" />);
+  const component = screen.getByTestId(testID);
+
+  expect(component.props.style).toStrictEqual({ color: "#00f" });
+
+  act(() => {
+    colorScheme.set("dark");
+  });
+  expect(component.props.style).toStrictEqual({ color: "#f00" });
+
+  act(() => {
+    colorScheme.set("unspecified");
+  });
+  expect(component.props.style).toStrictEqual({ color: "#00f" });
+});
--- a/src/__tests__/native/components.test.tsx
+++ b/src/__tests__/native/components.test.tsx
@@ -1,5 +1,6 @@
 import {
   Button as RNButton,
+  ScrollView as RNScrollView,
   TextInput as RNTextInput,
   type ButtonProps,
   type TextInputProps,
@@ -11,6 +12,7 @@ import { TextInput } from "react-native-css/components/TextInput";
 import { registerCSS, testID } from "react-native-css/jest";
 import { useCssElement } from "react-native-css/native";
 import type {
+  ComponentPropsDotNotation,
   StyledConfiguration,
   StyledProps,
 } from "react-native-css/runtime.types";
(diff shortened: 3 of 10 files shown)


YevheniiKotyrlo:
Cross-referencing #415 and #429 — they merge with **zero conflicts** and the result does not compile. Flagging it now, because nothing will warn whoever lands the second one.

`git merge-tree --write-tree` reports no conflict at all. Running `tsc --noEmit` on that merged tree gives exactly one error, repo-wide:

```
src/native/api.tsx(83,31): error TS2345: Argument of type 'ColorSchemeName' is not
  assignable to parameter of type '…Appearance").ColorSchemeName'.
  Type '"unspecified"' is not assignable to type 'ColorSchemeName'.
```

The two halves are each correct alone:

- **#429** replaces react-native's `ColorSchemeName` with this package's own union, which adds `"unspecified"`, and introduces a `setAppearanceColorScheme` seam in `src/color-scheme.ts` — routing `src/web/api.tsx` through it.
- **#415** adds `Appearance.setColorScheme(value)` at `src/native/api.tsx:83`, typed by whichever `ColorSchemeName` is in scope.

On the merge, #429's wider union reaches a call react-native 0.81 types as `'light' | 'dark' | null | undefined`, and `"unspecified"` is not a member.

**Neither PR can pre-empt this, which is why it is a note rather than a fix on one of them.** #415 cannot import `src/color-scheme.ts` — that file does not exist on `main` or on its branch. #429 has no `Appearance.setColorScheme` call in `src/native/api.tsx` to route, because on its branch that file has no setter.

The resolution is one line at merge time: route the native setter through the same seam the web one already uses, rather than calling `Appearance` directly.

```ts
// src/native/api.tsx
import { setAppearanceColorScheme } from "../color-scheme";
…
setAppearanceColorScheme(value);   // instead of Appearance.setColorScheme(value)
```

Measured: that takes the merged tree from one error to zero. The seam exists precisely because the one member the supported react-native range declares incompatibly is the "follow the system" write, so this is what it was built for.

Worth pairing with a compile-level guard, since the failure is a type error rather than a behaviour: the merged tree must typecheck. A runtime twin already exists on #429 at `src/__tests__/native/color-scheme.test.tsx:74-76`, which drives `colorScheme.set("unspecified")`.

YevheniiKotyrlo:
Following up on the cross-reference above with two measured additions, one of which is a trap worth naming explicitly.

## An inline cast at the call site compiles, and is the wrong resolution

The obvious shortcut for the merge break is to widen the argument in place rather than route through the seam:

```ts
Appearance.setColorScheme(value as Parameters<typeof Appearance.setColorScheme>[0]);
```

That does take the merged tree to zero type errors, on both react-native ends of the peer range. It is still wrong, because it silences **both** arms of a union that disagrees for a reason:

```
0.81.5  Appearance.js   NativeAppearance.setColorScheme(colorScheme ?? 'unspecified');
0.82.0  Appearance.js   NativeAppearance.setColorScheme(colorScheme);
```

Up to 0.81.5 a nullish request is coerced to `"unspecified"` before it reaches the native module. From 0.82.0 the coercion is gone and the argument is stored and forwarded verbatim. `colorScheme` is exported from `react-native-css/runtime`, and #429 deliberately widens the public `ColorScheme.set` to accept whichever spelling the installed version emits — so with a cast in place, `colorScheme.set(null)` puts `null` into `Appearance`'s cache and hands `null` to the native module on ≥0.82.

That is the failure #415's own comment forbids in the same function: *"Announcing it would put a value in Appearance's cache that no reader can render."* `setAppearanceColorScheme` exists precisely to prevent it, so the seam is the resolution and the cast routes around it.

## The boundary is 0.82.0, not 0.86

Measured across every react-native in the Yarn cache, by extracting `Libraries/Utilities/Appearance.d.ts` from each:

```
0.81.2 / 0.81.4 / 0.81.5   type ColorSchemeName = 'light' | 'dark' | null | undefined;
0.82.0 … 0.86.0            type ColorSchemeName = 'light' | 'dark' | 'unspecified';
```

The type narrowing and the loss of the `?? 'unspecified'` coercion land in the same release. #429 already names 0.82 on its own surface; #415 said 0.86 in four comments and now names 0.82 too, so the two branches no longer disagree about one boundary when they meet.

One nuance worth recording for whoever touches this next, because it is easy to collapse into the same number: there is a **third** state later in the range, where `setColorScheme` resolves `"unspecified"` against the OS before storing it, rather than storing it verbatim. It is present in 0.85.3 and absent in 0.84.1. I have not placed it more tightly than `(0.84.1, 0.85.3]`, so nothing above names it as a version.

## A latent one, for #415

#415's three new colour-scheme test files do not type-check against ≥0.82 — `colorScheme.set(null)` and friends, where `null` is no longer in `ColorSchemeName`. That includes `color-scheme-appearance-rn-0-86.test.tsx`, which models 0.86 while being written against 0.81's type surface. It is latent for both PRs because neither bumps `devDependencies.react-native` past 0.81.4, so CI never exercises the surface either PR exists to support — which is itself worth a look on #429.

PR #434, opened 2026-08-19
fix(native): one resolved-style cache key per distinct input set

### LINKED ISSUE #433: Per-instance config identity defeats the resolved-style cache (2307 entries where 417 suffice)
## Summary

`useCssElement` derives its style config **per component instance**, but the resolved-style cache is keyed on that config's **object identity**. Because every shipped `components/*.tsx` wrapper passes a module-constant `mapping`, N mounted elements produce N distinct config identities carrying identical values — so N identical elements each get their own `stylesFamily` entry, their own sorted rule array and their own observable, where one would do.

Measured on a real screen: **2307 cache entries where 417 suffice** (5.5×).

`styled()` does not have this problem — it derives the config once at module scope. Both call sites are in the same file, ten lines apart.

## Environment

- `react-native-css` **3.0.7** (current `latest`), and verified still present on `main`
- React Native 0.86, Fabric + Hermes, Expo SDK 57
- iPhone 12 Pro Max, iOS 26.6, **release** build

## Steps to reproduce

`stylesFamily` is not reachable from any public export (`./native` re-exports `api` only; `./native-internal` publishes `root` / `style-collection` / `variables`), so observing it needs a one-line instrumentation in the package's own source.

1. In `src/native/styles/index.ts`, after the family is created:

   ```ts
   export const stylesFamily = family((hash, rules) => { /* unchanged */ });

   // `family()` exposes `delete` / `keys` / `clear` and no `size`, so count via keys().
   setInterval(() => {
     console.log("stylesFamily entries=" + [...stylesFamily.keys()].length);
   }, 1500);
   ```

2. In any RN app built against that source, render many elements sharing one class list:

   ```tsx
   import { View } from "react-native"; // rewritten to components/View.tsx by the Metro resolver

   const Probe = () => (
     <>
       {Array.from({ length: 150 }, (_unused, index) => (
         <View key={index} className="size-5 rounded-full border bg-red-500" />
       ))}
     </>
   );
   ```

3. Read the count once the tree has mounted.

**Isolating comparison** — swap the rewritten `View` for `styled(RNView, { className: { target: "style" } })`, which derives its config once at module scope, and re-read the count on the identical tree. Nothing else changes: same class list, same rules, same element type.

## Expected result

The count reflects the number of **distinct class lists** on screen. 150 elements sharing one class list contribute one entry, because the cache exists to make resolution O(distinct class sets) rather than O(instances).

## Actual result

The count scales with the number of **instances**.

On a screen of 150 identical controls (four styled elements each, plus surrounding chrome):

| build | `stylesFamily` entries |
| --- | --- |
| `3.0.7` as published | **2307** |
| with `mappingToConfig` memoised on the mapping identity | **417** |

1890 redundant entries, each a separately sorted rule array and a retained observable over the same `calculateProps`, held for as long as the tree is mounted.

(417 rather than 1 is the honest floor for a real screen — it carries many distinct class lists, and e.g. `ScrollView` legitimately contributes two configs. What collapses is precisely the per-instance duplication.)

## Root cause

The cache key ([`src/native/react/rules.ts:300`](https://github.com/nativewind/react-native-css/blob/main/src/native/react/rules.ts#L300)) hashes the config by **object identity**, via a `weakFamily` that assigns a fresh number to each new object:

```ts
const keys = [state.configs, ...iterableKeys];
return generateHash(keys);
```

The rules half of that key shares correctly — `StyleCollection.styles(className)` is keyed by the class **string**.

The config half does not. In [`src/native/api.tsx`](https://github.com/nativewind/react-native-css/blob/main/src/native/api.tsx), the two entry points disagree:

```ts
// line 54 — styled(): once, at module scope. Shares.
const configs = mappingToConfig(mapping);

// line 90 — useCssElement(): once per INSTANCE. Does not share.
const [config] = useState(() => mappingToConfig(mapping));
```

and every shipped wrapper takes the second path with a module-constant mapping ([`src/components/View.tsx:11,18`](https://github.com/nativewind/react-native-css/blob/main/src/components/View.tsx#L11)):

```tsx
const mapping = { className: { target: "style", resetsTextAncestor: true } };
// ...
return useCssElement(RNView, props, mapping);
```

`mappingToConfig` is pure in `mapping`, so it returns the same **value** every time and a different **identity** every time. The cache is defeated by an identity that always carries the same value.

## Proposed fix

Memoise the derived config on the mapping identity — every call site the library itself ships already passes a stable module constant:

```ts
const MAPPING_CONFIG_CACHE = new WeakMap<object, Config[]>();

export function mappingToConfig(mapping) {
  const cached = MAPPING_CONFIG_CACHE.get(mapping);
  if (cached !== undefined) return cached;
  const configs = /* existing body */;
  MAPPING_CONFIG_CACHE.set(mapping, configs);
  return configs;
}
```

Notes:

- `useCssElement`'s `useState` can stay or go — with `mappingToConfig` memoised it returns the shared array either way.
- A caller passing a fresh mapping literal per render is unaffected (the `WeakMap` simply never hits), so this cannot make any call site worse.
- `state.configs` is read-only in its consumers (iterated for `source` / `target`), so sharing one array across components is safe.

## Impact, stated honestly

Two measurements, and they say different things:

- **Entry count: unambiguous and large** — 2307 → 417 on the screen above. The redundant work and the retained objects are real.
- **CPU: small in our workload** — one Instruments `--cpu` run per side over a 160-instance stress scene showed the JS thread at 2914ms → 2794ms and 2129ms → 2082ms, which is inside our suite's run-to-run spread. So we can't claim a CPU win from our own numbers.

That bounds what removing it buys *us*, and says nothing about a workload with more distinct class sets, more configs per element, or a slower device. Reporting the defect and both numbers rather than the flattering one — you're better placed to judge what it's worth on the workloads the library targets.

Happy to open a PR if the approach looks right.

YevheniiKotyrlo:
Fixes #433.

## The defect

`stylesFamily` is keyed by `generateStateHash`, which hashes weak keys by **object identity** — so the cache shares exactly when equal inputs arrive as one object:

```ts
// src/native/react/rules.ts
const stylesObs = stylesFamily(generateStateHash(state, rules), rules);
const keys = [state.configs, ...iterableKeys];
```

The rules half holds: `StyleCollection.styles(className)` is keyed by the class string. Three defects break the rest.

### 1. The config is minted per instance

```ts
// api.tsx:54 — styled(): once, at module scope. Shares.
const configs = mappingToConfig(mapping);
// api.tsx:90 — useCssElement(): once per INSTANCE. Does not.
const [config] = useState(() => mappingToConfig(mapping));
```

`mappingToConfig` is pure in `mapping`, so it returns the same **value** and a different **identity** every time — and every wrapper this library ships takes the second path with a module constant. N identical elements therefore hold N entries, each a separately sorted rule array and a retained observable. `getRuleVariation`'s rule clone is keyed on the same identity, so that was per element rather than per mapping.

### 2. `family` and `weakFamily` cache on truthiness

```ts
let value = map.get(key);
if (!value) { value = fn(key, args); map.set(key, value); }
```

`hashKeyFamily` returns `hashKeyCount++`, so the first weak key hashed is assigned `0`, reads back as a miss, and never settles:

```
generateHash([key]) x3, first object hashed:   "v"  "x"  "x"
generateHash([key]) x3, any later object:      "z"  "z"  "z"
```

### 3. The key is lossy, and a collision is silent

Folding every key's number into one number and printing it base-36 is a digest. `family` returns the entry it holds and **ignores the `rules` the second caller brought**, so two rule sets meeting on one string do not cost a cache miss — the second element renders the first element's styles.

`(a ^ b) + (a + 31)(b + 31)` is a coarse lattice over consecutive integers, so collisions concentrate where the library spends its time:

| key shape | current | this PR |
| --- | --- | --- |
| 1 config + 1 rule — a bare `<View className="text-sm" />` | **31.99%** (28792/90000) | 0 |
| all pairs over keys 0..2999 | 39.26% | 0 |

End to end on `main`: `.probe-red` and `.probe-blue` both hash to `4u1`, resolve to the **same observable**, and the red element renders `{"color":"#00f"}`. Over those 90000 inputs the fold held 61208 entries — 28792 of the "saved" entries were wrongly-rendered elements.

The three interact: fixing (1) and (2) re-numbers the keys and moves which pairs collide, so a fix leaving the key lossy trades a memory win for a rendering bug.

## The fix

**Derive one config per mapping**, through this library's own `weakFamily` — already the derive-once-per-identity primitive behind `hoverFamily`, `getRuleVariation` and `hashKeyFamily`. The function body is untouched.

**Cache on presence** — `if (!value)` becomes `if (value === undefined)`. Of the twelve `family`/`weakFamily` call sites, `hashKeyFamily` is the only one whose factory can return a falsy value, so this changes behaviour at exactly the broken site.

**Key injectively** — collect the numbers into a `Float64Array`, sort it, join. The typed array is what makes the ordering native: `Array.prototype.sort` needs a comparator to order numbers and calls it O(n log n) times, and on Hermes that callback is most of what the ordering costs (see Cost). Sorting preserves the order-independence the fold had, which the caller relies on because the rule set is a `Set` built in render order; joining rather than folding makes distinct key sets distinct strings. `MOD` and `PRIME` have no remaining reader.

The loop's inherited `if (!key) continue` goes with them. A skipped key is erased from the value, so `[a, b]` and `[a, falsy, b]` encoded to one string — the same collision class, surviving inside the rewrite meant to remove it. Nothing reachable is affected: `WeakKey` admits no falsy value, and all four key sources are objects by construction. What changes is that an out-of-domain key now fails at the `WeakMap`, rather than either throwing later in `calculateProps` or silently taking the colliding entry, depending on which of the two arrived at the cache first.

`generateStateHash`'s empty-string sentinel goes with it: a join makes that value collidable with a real state. It was unreachable — `state.configs` is always a key — and the two parameters no caller passes go at the same time.

`family` gains `size()` beside its existing `delete` and `clear`, because a cache whose size cannot be read is one whose central invariant cannot be regression-tested. `family` and `stylesFamily` are not themselves exported — but `rootVariables` and `universalVariables` are `family` instances published through `./native-internal`, so they gain the method too. Additive, and consistent with the `delete` and `clear` already on them.

## What it measures

50 identical elements, this repo's jest environment:

| configuration | entries |
| --- | --- |
| `main` | **50** |
| falsy-cache fix only | 50 |
| config derived once per mapping only | 2 |
| all three | **1** |

The residual `2` is the key `hashKeyFamily` assigns `0`.

A physical Android device (Xiaomi Mi MIX 3), 150 identical radio controls, reading `stylesFamily` through a temporary log: **1785 → 270-271** (the count settles a single entry apart between runs; 1785 is stable across all of them). The iOS measurement in the issue is 2307 → 417 on an iPhone 12 Pro Max; different screens, so the ratios are not comparable, but neither is a rounding effect.

### The cache stops growing without limit

`stylesFamily` is never released on unmount — `stylesObs.cleanup` is reached only when a component's observable CHANGES. That is a pre-existing defect and this PR does not fix it.

What changes is whether a second visit ADDS to what is held. A remount used to produce fresh config identities, fresh hashes and fresh entries while the old ones stayed:

| mount cycle | 1 | 2 | 3 | … |
| `main` | 2 | **4** | 6 | grows without limit |
| this PR | 2 | 2 | 2 | flat |

The bound moves from the number of MOUNTS — unbounded in a navigating app — to the number of distinct class lists, which is a property of the stylesheet.

### Rendered output is unchanged

Every element resolves the same declarations to the same values; what changes is how many times the library computes and retains them. The 1048 other tests in the suite, which assert rendered props extensively, are unchanged.

CPU is small and not claimed: one Instruments `--cpu` run per side over a 160-instance stress scene showed the JS thread at 2914ms → 2794ms and 2129ms → 2082ms, inside that suite's run-to-run spread.

## Is sharing one config safe?

While each element minted its own config, a write to one was private; sharing makes "the consumers only read it" load-bearing.

`Object.freeze` cannot check that here — a write to a frozen object throws only in strict mode, and in this repo's jest environment it does not throw at all, so a freeze-based test passes whatever the code does. The test compares the config's VALUE across a render instead, and is mutation-proven: injecting a write into `getStyledProps`'s config loop turns it red with the exact diff.

Three consequences:

- **An array `target` is aliased to the caller's array**, before and after. Deriving once changes how many config objects exist, never that aliasing.
- **A mutated mapping is no longer re-derived.** This narrows behaviour rather than changing it: `useState` already froze the derivation per instance, so a mutation reached only newly mounted elements — one mapping meaning two things depending on mount order.
- **A non-object mapping is refused by name.** `styled()` is reached from untyped JavaScript, where the type forbidding one is absent, and without the guard a primitive reaches the `WeakMap` and raises `Invalid value used as weak map key` from inside `reactivity` — naming neither `styled()` nor the argument. The predicate is deliberately narrower than that `WeakMap` contract: a function and an unregistered symbol are both valid weak keys, and both are refused, because neither is a mapping.

  It narrows six input kinds, which is worth your judgement rather than mine. `Object.entries` threw only for `null` and `undefined`; a number, boolean, function or symbol derived an empty config, and a string derived one config per character. None of those rendered anything usable — an empty config drops every prop, though note the type-legal `styled(C, {})` does that too and this PR does not change it. Two consequences follow: on the `styled()` path the throw lands at module evaluation rather than at render, and web's `styled` still no-ops on all six, so the platforms now disagree for input the types forbid.

## Cost

Instrumented over a 300-element render the key count is `min 5 · p50 10 · p90 15 · max 16`, so the band that matters is 5 to 16 keys.

The numbers below are **ratios between engines measured inside one run, not absolute timings**, captured on a physical Android device through a dev client whose clocks are thermally clamped. A uniform clamp scales every candidate in a run by the same factor, which is checkable rather than assumed: across two independent captures the ratios agreed to within 5.8% everywhere and 1.5% at 11 keys and above, while the absolutes they were computed from differed by more.

Warm lookups — hash plus `Map.get` — relative to the fold this replaces:

| keys | fold | `sort((a,b)=>a-b)` | `Float64Array.sort()` |
| --- | --- | --- | --- |
| 5 | 1.00 | 1.33 | 1.22 |
| 11 | 1.00 | 1.64 | **1.10** |
| 20 | 1.00 | 2.20 | 1.19 |
| 50 | 1.00 | 2.75 | 1.27 |

**The cost is the comparator, not the sort and not the string.** `Array.prototype.sort` cannot order numbers without one, and that comparator is a JS function the engine calls O(n log n) times. A typed array sorts numerically in native code with no callback at all. Isolating the ordering step on the same device, at 11 keys: `sort((a,b)=>a-b)` 7354 ns, `Float64Array.sort()` 3319 ns, no ordering at all 195 ns — so the callback is most of what the ordering costs, and removing it recovers most of that.

This is also where the engines disagree, which is worth stating because it decides the shape rather than merely the constant. On V8 the difference barely registers and an insertion sort is fastest at these sizes; on Hermes the same insertion sort loses badly by 50 keys, and the typed array wins at every size. The version above is the one that is never wrong on either.

`Float64Array` rather than `Int32Array`: `hashKeyCount` is unbounded and an int32 wraps silently at 2^31, which would map two distinct key sets onto one string — reintroducing exactly the collision this PR removes, in a form that only appears after two billion keys. Verified across 20 000 distinct key sets that the typed encoding agrees with the comparator sort on every one, stays order-independent, and collides on none.

**Order-independence buys nothing measurable on the screens I have.** Keying without the sort produced 271 distinct entries against the sorted key's 271 on a 150-control screen, and 119 against 119 on the shell alone. Those are counts, so no clock is involved. It is also not correctness — `"p-4 bg-red"` and `"bg-red p-4"` reach the same rules, so an order-dependent key costs a duplicate entry rather than wrong styles. I kept the sort because "no two class lists on these two screens were permutations of each other" is a fact about my stylesheet and not about your users'; at 1.10x it is no longer a trade worth making.
## Tests

Nineteen across five files, each written red first.

- `style-cache-sharing.test.tsx` — 50 identical elements produce one entry; different class lists keep distinct entries.
- `config-identity.test.ts` — two consumers of one mapping produce one key; differing mappings do not; a state hash is never the empty string; a non-object mapping is refused; a mutated mapping is not re-derived.
- `family.test.ts` — the factory-once-per-key contract for both families with a falsy result.
- `state-hash.test.ts` — distinct key sets never share a key; a key set encodes the same however ordered; a key outside `WeakKey` is refused rather than erased; the encoding is integers rather than floats or exponents.
- `style-cache-bounds.test.tsx` — nothing writes to the shared config; a shared entry survives being read; 25 mount cycles hold at one count.

Reverting each fix sends exactly the expected tests red:

| state | failing | entries for 50 elements | counts over 25 cycles |
| --- | --- | --- | --- |
| all three | 0 of 19 | 1 | 1 distinct |
| injective key reverted | 1 | 1 | 1 distinct |
| + config memo reverted | 4 | 50 | **25 distinct** |
| all three reverted | 7 | 50 | 25 distinct |

## Verification

`yarn typecheck` and `yarn lint` both exit 0. Full suite (`yarn test`), same runner, both sides measured together:

| | base (`f70c402`) | this branch |
| --- | --- | --- |
| passed | 1048 | **1067** |
| failed | 3 | 3 |
| skipped | 21 | 21 |

Exactly `+19`, and all nineteen pass. The 3 failures are identical on the unmodified base: they are the babel import-plugin's relative-import cases (`react-native › 7`, `react-native-web › 6`, `react-native-web › 17`), which are the Windows backslash-vs-forward-slash path defect I reported in #390 and are green on Linux and macOS. Nothing outside `src/__tests__/babel/` fails on either side.

One correction to the issue: its reproduction counts entries via `stylesFamily.keys()`, which does not exist on `main` — `family()` exposes `delete` and `clear` only. The tests above need no instrumentation.

### DIFF
--- a/src/__tests__/native/config-identity.test.ts
+++ b/src/__tests__/native/config-identity.test.ts
@@ -1,0 +1,121 @@
+import { ScrollView as RNScrollView, View as RNView } from "react-native";
+
+import { generateStateHash } from "../../native/react/rules";
+import {
+  mappingToConfig,
+  type ComponentState,
+  type Config,
+} from "../../native/react/useNativeCss";
+import {
+  VAR_SYMBOL,
+  type Effect,
+  type VariableContextValue,
+} from "../../native/reactivity";
+import type { StyledConfiguration } from "../../runtime.types";
+
+/**
+ * The resolved-style cache is keyed by `generateStateHash`, which hashes `state.configs` by object
+ * identity. Sharing therefore depends on one mapping producing one config array, rather than an
+ * equal array per consumer.
+ *
+ * The rules half of that key already holds: `StyleCollection.styles(className)` is keyed by the
+ * class string, so every element carrying the same class list references one rule set. The config
+ * half is the one input that is minted per consumer.
+ */
+
+/** A stand-in for the rule set a shared class list resolves to — any object is a valid weak key. */
+const ruleFor = (): WeakKey => ({});
+
+/** The smallest `ComponentState` `generateStateHash` reads: it only touches `configs`. */
+const stateFor = (configs: Config[]): ComponentState => {
+  const observers = new Set<Effect>();
+  const ruleEffect: Effect = { observers, run: () => undefined };
+  const inheritedVariables: VariableContextValue = { [VAR_SYMBOL]: true };
+
+  return {
+    configs,
+    inheritedContainers: {},
+    inheritedVariables,
+    ruleEffect,
+    ruleEffectGetter: (observable) => observable.get(ruleEffect),
+    styleEffect: { observers, run: () => undefined },
+  };
+};
+
+const viewMapping = {
+  className: "style",
+} satisfies StyledConfiguration<typeof RNView>;
+
+const scrollViewMapping = {
+  className: "style",
+  contentContainerClassName: "contentContainerStyle",
+} satisfies StyledConfiguration<typeof RNScrollView>;
+
+test("two consumers of one mapping share a cache key", () => {
+  const rules = [ruleFor()];
+
+  const first = generateStateHash(
+    stateFor(mappingToConfig(viewMapping)),
+    rules,
+  );
+  const second = generateStateHash(
+    stateFor(mappingToConfig(viewMapping)),
+    rules,
+  );
+
+  expect(first).toBe(second);
+});
+
+test("mappings that differ keep different cache keys", () => {
+  const rules = [ruleFor()];
+
+  const view = generateStateHash(stateFor(mappingToConfig(viewMapping)), rules);
+  const scrollView = generateStateHash(
+    stateFor(mappingToConfig(scrollViewMapping)),
+    rules,
+  );
+
+  expect(view).not.toBe(scrollView);
+});
+
+test("a state hash always carries the config, so it is never the empty string", () => {
+  // The empty string used to double as a no-keys sentinel in `generateStateHash`. With the key a
+  // join rather than a digest, an empty key list renders as the empty string too — so the sentinel
+  // and a real state would have shared one cache entry. The config is an unconditional key, which
+  // is what makes the sentinel unnecessary rather than merely unlikely.
+  const state = stateFor(mappingToConfig(viewMapping));
+
+  expect(generateStateHash(state, [])).not.toBe("");
+  expect(generateStateHash(state, [ruleFor()])).not.toBe("");
+});
+
+test("a mapping that is not an object is refused by name", () => {
+  // The derivation is cached on the mapping OBJECT. A primitive cannot be a weak-map key, so
+  // without this guard the failure surfaces as a `WeakMap` error naming nothing the caller wrote.
+  expect(() => mappingToConfig("style" as never)).toThrow(
+    /mapping must be an object/u,
+  );
+  expect(() => mappingToConfig(undefined as never)).toThrow(
+    /mapping must be an object/u,
+  );
+});
+
+test("a mapping mutated after its first use is not re-derived", () => {
+  // Documented rather than defended: `useCssElement` already froze the derivation per instance, so
+  // a mutation only ever reached NEWLY mounted elements — the same mapping meaning two things at
+  // once. Deriving once per mapping settles it on one.
+  const mapping: Record<string, string> = { className: "style" };
+  const first = mappingToConfig(mapping);
+
+  mapping.className = "contentContainerStyle";
+
+  expect(mappingToConfig(mapping)).toBe(first);
+});
+
+test("a mapping built per call still produces an equal config", () => {
+  // Every wrapper the library ships passes a module constant, but a caller may build the mapping
+  // inline. That path cannot share on identity and has to keep working unchanged.
+  expect(mappingToConfig({ className: "style" })).toEqual(
+    mappingToConfig(viewMapping),
+  );
+});
--- a/src/__tests__/native/family.test.ts
+++ b/src/__tests__/native/family.test.ts
@@ -1,0 +1,42 @@
+import { generateHash } from "../../native/react/rules";
+import { family, weakFamily } from "../../native/reactivity";
+
+/**
+ * `family` and `weakFamily` promise one factory call per key, and cached on the result being
+ * truthy. A factory that legitimately returns `0`, `""` or `false` was therefore re-run on every
+ * lookup, and its key never settled on a value.
+ *
+ * `hashKeyFamily` in `native/react/rules.ts` is such a factory: it hands out `hashKeyCount++`, so
+ * the first weak key ever hashed is assigned `0` and is the one key that never caches. Its hash
+ * changes between lookups, which splits every cache keyed on that hash.
+ */
+
+test("weakFamily calls its factory once per key when the result is falsy", () => {
+  let calls = 0;
+  const numbers = weakFamily<object, number>(() => calls++);
+  const key = {};
+
+  expect(numbers(key)).toBe(0);
+  expect(numbers(key)).toBe(0);
+  expect(calls).toBe(1);
+});
+
+test("family calls its factory once per key when the result is falsy", () => {
+  let calls = 0;
+  const numbers = family<string, number>(() => calls++);
+
+  expect(numbers("key")).toBe(0);
+  expect(numbers("key")).toBe(0);
+  expect(calls).toBe(1);
+});
+
+test("a weak key hashes to the same value on every lookup", () => {
+  // The key assigned `0` is the one the falsy-cache miss exposes, and it is whichever key this
+  // module hashes FIRST. Asserting the value rather than only the agreement is what keeps that
+  // true: a test added above this one that hashes would take the `0` and leave this passing
+  // against a key that was never at risk.
+  const key = {};
+
+  expect(generateHash([key])).toBe("0");
+  expect(generateHash([key])).toBe(generateHash([key]));
+});
--- a/src/__tests__/native/state-hash.test.ts
+++ b/src/__tests__/native/state-hash.test.ts
@@ -1,0 +1,73 @@
+import { generateHash } from "../../native/react/rules";
+
+/**
+ * `generateHash` keys the resolved-style cache. Two different sets of rules landing on one key does
+ * not cost a cache miss — `family` returns the entry it already holds and ignores the rules the
+ * second caller brought, so that element renders the first element's styles.
+ *
+ * Measured before this was injective: a `vertical-align: top` element rendered `object-fit: contain`
+ * because both rule sets hashed to `18h`.
+ */
+
+/** Distinct weak keys. Any object is a valid one. */
+const keysOf = (count: number): WeakKey[] =>
+  Array.from({ length: count }, () => ({}));
+
+test("distinct key sets never share a hash", () => {
+  const keys = keysOf(60);
+  const owner = new Map<string, string>();
+
+  for (const [leftIndex, left] of keys.entries()) {
+    for (const [offset, right] of keys.slice(leftIndex + 1).entries()) {
+      const signature = `${String(leftIndex)}+${String(leftIndex + 1 + offset)}`;
+      const hash = generateHash([left, right]);
+      const prior = owner.get(hash);
+
+      expect(prior ?? signature).toBe(signature);
+      owner.set(hash, signature);
+    }
+  }
+
+  // Vacuity guard: the loop above must actually have hashed every pair.
+  expect(owner.size).toBe((keys.length * (keys.length - 1)) / 2);
+});
+
+test("a key set hashes the same however it is ordered", () => {
+  // The rule set reaching `generateStateHash` is a Set built in render order, so order-independence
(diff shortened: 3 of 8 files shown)


YevheniiKotyrlo:
### Why this PR carries no device screenshot, and what the measurement is instead

The defect changes how much work happens, not what is painted — so a before/after pair would be two identical frames, which reads as evidence while proving nothing.

**The unambiguous number is the cache-entry count.** On a 160-instance stress screen, resolved-style entries went **2307 → 417** with this change applied. That is the redundant work and the retained objects, and it is large.

**The CPU number is honest and small.** iPhone 12 Pro Max release build, one `--cpu` Instruments run per side:

| scene | JS thread | main thread |
| --- | --- | --- |
| 160 switches, before | 2914ms | 1194ms |
| 160 switches, memoised | 2794ms | 1147ms |
| 160 checkboxes, before | 2129ms | 928ms |
| 160 checkboxes, memoised | 2082ms | 817ms |

One run per side against a suite whose run-to-run spread is wider than the difference — so I'm reporting this as "small, possibly nothing" rather than a percentage that would flatter it.

The two say different things and both belong here. The entry count bounds what the defect *is*; the CPU share bounds what removing it buys **my** workload, and says nothing about one with more distinct class sets, more configs per element, or a slower device.

Issue #445, opened 2026-09-09
exports: ./components/* declares .tsx source targets for two modules authored .ts/.cts — ERR_MODULE_NOT_FOUND under source and react-native


YevheniiKotyrlo:
## Summary

The `./components/*` wildcard in `package.json` declares `source` and `react-native` as `./src/components/*.tsx`, but two modules matched by that pattern are authored `.ts` / `.cts`. Under either condition those subpaths are a hard `ERR_MODULE_NOT_FOUND` — export conditions match on **key presence**, not target existence, so there is no fallthrough to `import`/`require`.

`react-native` is not a dormant condition: Expo sets it for iOS and Android (`unstable_conditionsByPlatform` in `@expo/metro-config`), and `source` is the condition `react-native-builder-bob`'s example-app workflow runs on.

## Affected

```
exports["./components/*"] [react-native] -> ./src/components/copyComponentProperties.tsx
exports["./components/*"] [react-native] -> ./src/components/index.tsx
exports["./components/*"] [source]       -> ./src/components/copyComponentProperties.tsx
exports["./components/*"] [source]       -> ./src/components/index.tsx
```

Authored files, from the published tarball:

```
$ ls src/components/ | grep -E '^(index|copyComponentProperties)\.'
copyComponentProperties.ts
index.cts
index.ts
```

Reproduced against the published `react-native-css@3.0.7`:

```
$ node --conditions=source -e "…require.resolve(…)"
[source] react-native-css/components/copyComponentProperties -> FAILED: MODULE_NOT_FOUND
```

Note that the exact `./components` entry already answers `index` with `./src/components/index.cts`, so the wildcard's `.tsx` guess contradicts the map's own neighbouring entry.

## Suggested fix

An exact key beats a pattern regardless of declaration order, so two entries alongside the existing `./components/react-native-gesture-handler` resolve it:

```jsonc
"./components/copyComponentProperties": {
  "source": "./src/components/copyComponentProperties.ts",
  "react-native": "./src/components/copyComponentProperties.ts",
  "import":  { "types": "./dist/typescript/module/src/components/copyComponentProperties.d.ts",   "default": "./dist/module/components/copyComponentProperties.js" },
  "require": { "types": "./dist/typescript/commonjs/src/components/copyComponentProperties.d.ts", "default": "./dist/commonjs/components/copyComponentProperties.js" }
},
"./components/index": {
  "source": "./src/components/index.cts",
  "react-native": "./src/components/index.cts",
[...]

PR #446, opened 2026-09-09
fix(exports): correct `./components/*` source targets for two non-tsx modules

YevheniiKotyrlo:
Fixes the four broken targets reported in #445.

## The defect

`./components/*` declares `source` and `react-native` as `./src/components/*.tsx`, but two modules the pattern matches are not authored `.tsx`:

```
$ ls src/components/ | grep -E '^(index|copyComponentProperties)\.'
copyComponentProperties.ts
index.cts
index.ts
```

Export conditions match on **key presence**, not target existence, so there is no fallthrough to `import`/`require` — both subpaths are a hard `ERR_MODULE_NOT_FOUND` under either condition:

```
$ node --conditions=source -e "…require.resolve(…)"
[source] react-native-css/components/copyComponentProperties -> FAILED: MODULE_NOT_FOUND
[source] react-native-css/components/index                   -> FAILED: MODULE_NOT_FOUND
```

`react-native` is not dormant — Expo sets it for iOS and Android via `unstable_conditionsByPlatform`. `source` is the condition `react-native-builder-bob`'s example-app workflow runs on.

## The fix

Two exact keys. An exact key beats a pattern regardless of declaration order, mirroring the existing `./components/react-native-gesture-handler` entry.

`index` takes `.cts` rather than `.ts` because that is what the exact `./components` entry already declares — the wildcard's `.tsx` guess contradicted the map's own neighbouring entry.

## Verification

Applied against the published 3.0.7 tarball, both subpaths resolve under `source` and `react-native`, and every other entry is unchanged. I checked the whole map rather than just these two: **509 declared targets across every entry and every condition, 0 missing** after the fix.

That sweep is how the defect surfaced. It walks the real `dist/module/**` tree, expands each wildcard over it (skipping stems that have an exact key, since exact beats pattern) and asserts every declared target exists. It is generic — nothing in it is specific to my project — and it needs `dist/**`, so it belongs as a post-`prepare` check rather than an ordinary jest test. Happy to add it to this PR if you would like it; I left it out to keep this reviewable as a one-file fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,30 @@
         "types": "./dist/typescript/commonjs/src/components/index.d.ts"
       }
     },
+    "./components/copyComponentProperties": {
+      "source": "./src/components/copyComponentProperties.ts",
+      "react-native": "./src/components/copyComponentProperties.ts",
+      "import": {
+        "types": "./dist/typescript/module/src/components/copyComponentProperties.d.ts",
+        "default": "./dist/module/components/copyComponentProperties.js"
+      },
+      "require": {
+        "types": "./dist/typescript/commonjs/src/components/copyComponentProperties.d.ts",
+        "default": "./dist/commonjs/components/copyComponentProperties.js"
+      }
+    },
+    "./components/index": {
+      "source": "./src/components/index.cts",
+      "react-native": "./src/components/index.cts",
+      "import": {
+        "types": "./dist/typescript/module/src/components/index.d.ts",
+        "default": "./dist/module/components/index.cjs"
+      },
+      "require": {
+        "types": "./dist/typescript/commonjs/src/components/index.d.ts",
+        "default": "./dist/commonjs/components/index.cjs"
+      }
+    },
     "./components/react-native-safe-area-context": {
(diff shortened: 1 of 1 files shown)


YevheniiKotyrlo:
### Why this PR carries no device screenshot, and no test

Both absences have the same cause: this changes two `exports` targets in `package.json`, so what it fixes is **module resolution**, which happens before any frame is painted and outside anything jest can observe.

The measurement is in the body and it is exhaustive rather than illustrative: **509 declared targets across every entry and every condition, 0 missing** after the fix. Before it, two are a hard `ERR_MODULE_NOT_FOUND` under either `source` or `react-native`, because export conditions match on key presence rather than target existence — so there is no fallthrough to `import`/`require` to rescue them.

The sweep that produced that number walks the real `dist/module/**` tree and expands each wildcard over it, skipping stems that have an exact key since exact beats pattern. It needs `dist/**`, so it belongs as a post-`prepare` check rather than an ordinary test; I left it out to keep this reviewable as a one-file fix and am happy to add it.

PR #449, opened 2026-09-10
fix(compiler): drop a namespace-qualified attribute selector

YevheniiKotyrlo:
## Summary

`component.namespace` is discarded, so a namespace-qualified attribute selector matches the **unqualified** set instead of nothing — including an undeclared prefix, which Selectors L3 §6.3.3 makes an invalid selector.

## Problem

Measured — all five forms compile to the identical query `["d", "x", "=", "a"]`:

| selector | Selectors §6 | before |
| --- | --- | --- |
| `[data-x='a']` | attributes in no namespace | matches — correct |
| `[\|data-x='a']` | the same set, explicitly | matches — correct |
| `[*\|data-x='a']` | attributes in **any** namespace | matches — correct here, see below |
| `[ns\|data-x='a']` with `@namespace ns …` | only attributes in `ns` | **matches** |
| `[undeclared\|data-x='a']` | §6.3.3 — invalid, so the selector is dropped | **matches** |

The last row is the worse half: a rule the author's own stylesheet says cannot apply was applying. lightningcss reports the prefix rather than rejecting it, so dropping it is this compiler's job.

## Solution

A specific namespace can never match, because no React Native prop carries one — so the selector is dropped, at both attribute build sites.

The first three rows are untouched and stay correct for a reason worth stating: with no namespaced props in the tree, "no namespace" and "any namespace" denote the same set, so `[*|att]` needs no special handling.

```ts
function isNamespacedAttribute(component): boolean {
  return component.namespace?.type === "specific";
}
```

## Test plan

Five cases in `src/__tests__/native/attributes.test.tsx` — the three that must keep matching, plus a declared and an undeclared prefix. Mutation-proved: making the predicate `false` turns exactly the last two red.

`yarn typecheck` and `yarn lint` clean; `yarn test` 1053 passed, 3 failed, 21 skipped. The three failures are two babel suites that fail identically on an untouched `main` worktree (Windows-only module-specifier rewrites) — this touches no babel file.

## Note

Independent of #447 and #448, though all three touch the same two build sites — whichever lands last needs a trivial rebase.

### DIFF
--- a/src/__tests__/native/attributes.test.tsx
+++ b/src/__tests__/native/attributes.test.tsx
@@ -133,3 +133,43 @@ describe("dataSet attribute selector", () => {
     });
   });
 });
+
+describe("a namespace-qualified attribute selector represents nothing", () => {
+  const matchedWidth = (selector: string): number | undefined => {
+    registerCSS(`.test${selector} { width: 10px; }`);
+    render(
+      <Text testID={testID} className="test" {...{ dataSet: { x: "a" } }} />,
+    );
+    const style = screen.getByTestId(testID).props.style as
+      | { width?: number }
+      | undefined;
+    return style?.width;
+  };
+
+  // Measured against lightningcss: only `[ns|att]` reports a `specific` namespace.
+  // `[att]` and `[|att]` report none and `[*|att]` reports `any`, and with no
+  // namespaced props in the tree those three denote the same set.
+  test.each([
+    ["no namespace", `[data-x='a']`],
+    ["explicitly no namespace", `[|data-x='a']`],
+    ["any namespace", `[*|data-x='a']`],
+  ])("%s matches the prop", (_label, selector) => {
+    expect(matchedWidth(selector)).toBe(10);
+  });
+
+  test("a declared prefix matches nothing — no prop is in a namespace", () => {
+    registerCSS(
+      `@namespace ns url(http://example.com/ns); .test[ns|data-x='a'] { width: 10px; }`,
+    );
+    render(
+      <Text testID={testID} className="test" {...{ dataSet: { x: "a" } }} />,
+    );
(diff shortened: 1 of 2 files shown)


YevheniiKotyrlo:

**UNFIXED —** Both attributed bars are red. The namespace qualifier is discarded, so a selector Selectors L3 §6.3.3 makes INVALID matches the unqualified set instead of nothing.

**FIXED —** Only the unqualified bar is red. A prefixed attribute selector represents nothing, because no React Native prop carries a namespace.

| --- | --- |
|  |  |

**This is the over-match direction, so the UNFIXED frame is the one with MORE paint on it.** Worth saying before you read the pair: a dropped declaration is visible the first time anyone looks at a screen, and this one paints — on elements the author's own stylesheet says it cannot reach.

Two rules over one attribute. The first bar is unqualified and must keep matching on both builds — it is the control, and without it two blue bars on the right would be indistinguishable from a build that dropped both rules. The second is the same selector behind an **undeclared** prefix over an element that does carry the attribute. The third carries the qualified rule's class and no attribute at all, so the withheld case has something to be compared to.

The undeclared arm rather than a declared one, because `@namespace` is valid only before any style rule and this probe's stylesheet is imported beneath others — and because it is the worse half: a declared prefix merely over-matches, while an undeclared one applies a selector the stylesheet has already invalidated. lightningcss reports the prefix rather than rejecting it, which is what leaves the drop to this compiler.

Both frames come from the same device in the same run (Android 36 emulator, 1140×2400 @ 480dpi), but `before` is stock `3.0.7` rather than "this build minus this PR" — so one unrelated difference is visible and worth naming: the compiler's `inlineRem` defaults to `14` and our build sets it to `16`, so every rem-derived length in the `before` frame renders at 87.5% of the `after` one. That is a different fix, not this one.

